PR #1, opened 2023-12-20
DEVEXP-233: Update Numbers API from latest specification

asein-sinch:
**FEATURES:**
 - Regeneration of the models from the latest OAS document
   - Available regions types becomes an array
   - The callbackConfiguration API becomes the callbacks API (breaking change in the SDK)
   - ActiveNumberRequest has a new optional parameter: `callbackUrl `
   - RentAnyNumberRequest has a new optional parameter: `callbackUrl `
 - the API client defines a new optional boolean parameter `repeatParamArray` in order to manage the APIs that request as many query parameters as there are elements in the array.

**BUG FIXES**
 - samples: update the configuration to add a parameter `SMS_REGION`


packages/numbers/src/models/v1/callback-payload/callback-payload.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ */
+
+
+export interface CallbackPayload {
+
+  /** The ID of the event. */
+  eventId?: string;
+  /** The date and time when the callback was created and added to the callbacks queue. */
+  timestamp?: Date;
```
JPPortier, line 14: Because we known the current payload do not contains TZ but fix will be delivered in the future; is translation from String to Date is correctly handled here  ?

asein-sinch: For the moment no, I'm waiting for my account to be reset in order to test and see what is wrong in the response. I'll create a response plugin to "patch" the response before it's returned to the end user.
But maybe you can help me here an tell me which part is missing in the expected response: it should be ISO 8601 but the part saying it is zulu time is missing, right? If it is the case, the data will still be transformed into a Date object but with a wrong offset.

JPPortier: Sample of received payload:
```
{
    "eventId": "XXXX",
    "timestamp": "2023-12-22T09:35:38.53207693",
    "projectId": "XXXX",
    "resourceId": "+17813341312",
    "resourceType": "ACTIVE_NUMBER",
    "eventType": "PROVISIONING_TO_VOICE_PLATFORM",
    "status": "SUCCEEDED",
    "failureCode": null
}
```
The `timestamp` value do not contains time zone information: `2023-12-22T09:35:38.53207693`
For Java is was causing issue; so comment is just to ensure there is issue here when providing a Date from the JSON

asein-sinch: Thanks, it's clear now.
I'll provide a helper function that SDK user will have to use after parsing the event. This function will check is the property `timestamp` contains a timezone and add `Z` if not the case

JPPortier: So the `timestamp` field will be defined as string within `CallbackPayload` and no longer as `Date` ?

asein-sinch: No, the interface stays the same. I've added a new commit where I provide a helper method to parse the event body

packages/numbers/src/rest/v1/callbacks/callbacks-webhook.ts
```diff
@@ -0,0 +1,15 @@
+import { CallbackPayload } from '../../../models';
+
+export const parseEventNotification = (eventBody: any): CallbackPayload => {
+  // There is a bug in the API which doesn't send the timezone along with the timestamp
+  // As the server formats the timestamp as GMT, we check if the timezone is missing and add it if needed
```
JPPortier, line 5: ❤️

JPPortier (approved):

PR #2, opened 2023-12-27
DEVEXP-225: Break circular dependencies

asein-sinch:
There was some circular dependencies between the imports and the API client due to the fact the OAuth2 plugin needs to use the API client and the API client needs to use the OAuth2 plugin when managing the token refresh.
This pull request reorganise the imports / exports in order to have a finer control on what is exposed and remove the circular dependencies


packages/sdk-client/src/client/api-client-helpers.ts
```diff
@@ -51,10 +50,10 @@ export async function invalidateAndRegenerateJwt(
   errorContext: ErrorContext,
 ): Promise<RequestOptions> {
   const oauth2Plugin = requestPlugins?.find(
-    (plugin) => plugin instanceof Oauth2TokenRequest,
-  ) as Oauth2TokenRequest;
+    (plugin) => plugin.getName() === 'Oauth2TokenRequest',
```
JPPortier, line 53: Can we imagine having constants shared by plugin to avoid hard coded string comparison?
(in the future in case any change from plugin's name; code will be broken)

asein-sinch: Sure thing! I've updated the code in a new commit

JPPortier (approved):

PR #7, opened 2023-12-28
DEVEXP-252: Voice API - Write examples

asein-sinch:


packages/voice/src/rest/v1/voice-api.ts
```diff
@@ -93,7 +93,12 @@ export class VoiceApi implements Api {
   }
 
   private buildBasePath(region: VoiceRegion) {
-    return `https://calling${region}.api.sinch.com`;
+    switch (this.apiName) {
+    case 'ApplicationsApi':
+      return `https://callingapi.sinch.com`;
```
JPPortier, line 98: https://calling.api.sinch.com with a dot between `calling` and `api`? (for default region?) 

For my knowledge, what is context for  `this.apiName` = `ApplicationsApi` and why is `this.apiName` is having effect onto region to be used ?

asein-sinch: The Voice API defines the URL `https://calling{region}.api.sinch.com` for all endpoints except for the following paths:
 - /v1/configuration/numbers (GET / POST / DELETE) to list / assign / unassign the numbers with applications
 - /v1/configuration/callbacks/applications/{applicationkey} (GET / POST) to retrieve / update the callback urls
 - /v1/calling/query/number/{number} (GET) to retrieve information about a number
 For these 6 resources, the `servers` property is overridden with `https://callingapi.sinch.com`
 
 `this.apiName` is the value computed by the TS generator with the tag information: here `Applications`

JPPortier (approved):

PR #8, opened 2024-01-09
DEVEXP-256: Add timezone response plugin

asein-sinch:
This new response plugin is (hopefully) a temporary hack to add the missing timezone (assumption taken it's `Z`) for the API response where it's missing.

### DIFF
--- a/packages/sdk-client/tests/plugins/timezone/timezone.response.test.ts
+++ b/packages/sdk-client/tests/plugins/timezone/timezone.response.test.ts
@@ -1,0 +1,45 @@
+import { TimezoneResponse } from '../../../src';
+import { ResponsePluginContext } from '../../../src/plugins/core/response-plugin';
+import { Headers } from 'node-fetch';
+
+describe('Timezone response plugin', () => {
+
+  let context: ResponsePluginContext;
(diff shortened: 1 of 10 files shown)


packages/sdk-client/tests/plugins/timezone/timezone.response.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const apiResponse = {
+      timestamp: TIMESTAMP_WITH_MISSING_TIMEZONE,
+    };
+    const plugin = new TimezoneResponse();
+    const runner = plugin.load(context);
+    const result = await runner.transform(apiResponse);
+
+    expect(result.timestamp).toBe(TIMESTAMP_WITH_MISSING_TIMEZONE);
+  });
+});
```
Dovchik, line 45: What happens when API is updated but SDK is not, 
Can you check for case when correct timezone is returned in response?

asein-sinch: New test added

JPPortier: Not required but another test with a `+XX` timezone format could ensure regexp coverage

JPPortier (approved):

PR #10, opened 2024-01-19
DEVEXP-269: Fix webhooks

asein-sinch:
- Create sample app for webhooks
 - Add validation for signed request (and basic auth in case it happens in webhooks)
 - Add new models for voice webhooks responses + fix some models according to inconsistencies viewed during manual tests


packages/sdk-client/src/utils/authorization.helper.ts
```diff
@@ -0,0 +1,151 @@
...
+  normalizedHeaders: {[p: string]: string | string[]},
+  applicationKey: string,
+  applicationSecret: string,
+  path: string,
+  body: any,
+  method: string,
+): boolean => {
+  const authKeyAndSecret = authorizationValue.split(':');
+  if(authKeyAndSecret.length !== 2) {
+    throw new Error('Invalid authorization value format provided');
```
JPPortier, line 98: From a "consumer" point of view, is it better to
- return `false` from `validateAuthenticationHeader` call because of invalid authentication ?
- or throw an exception ?

Currently the exception is not dedicated to an auth issue.
So even if catched from application/consumer; it won't enable to identified an auth issue vs another one.

Without real useCase, my personal feeling is to let final application handle not-authentificated feature itself by:
- adding here a log about the error
- return false without exception... 

but a `documented` and `dedicated` Exception could do the job too...

packages/voice/src/rest/v1/applications/applications-api.ts
```diff
@@ -62,7 +62,7 @@ export class ApplicationsApi extends VoiceApi {
     };
 
     const body: RequestBody = '';
-    const path = `/v1/calling/query/number/${data['number']}`;
+    const path = `/v1/calling/query/number/${data['number'].split(" ").join("")}`;
```
JPPortier, line 65: May be a comment about current server behavior explaining 'why' we have to modify input with such "trick".
For future generation education purpose or even ourselves in few weeks/months 🥲

examples/simple-examples/src/voice/callouts/custom.ts
```diff
@@ -24,12 +24,50 @@ import { CustomCalloutRequestData, VoiceRegion } from '@sinch/sdk-core';
...
+              },
+              {
+                id: 'confirm',
+                mainPrompt: '#tts[Thank you for confirming your order. Enter your 4-digit PIN.]',
+                maxDigits: 4,
+              },
+            ],
+          },
+        } as AceResponse),
+        pie: "https://21cb-92-95-245-242.ngrok-free.app",
```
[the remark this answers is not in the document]
asein-sinch: Ooopsy

examples/webhooks/src/controller/app.controller.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  constructor(
+    private readonly numbersService: NumbersService,
+    private readonly smsService: SmsService,
+    private readonly verificationService: VerificationService,
+    private readonly voiceService: VoiceService) {}
+
+  @Post('/numbers')
+  public numbers(@Body() eventBody: any, @Res() res: Response) {
+    // console.log(eventBody);
```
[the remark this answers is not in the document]
asein-sinch: This comment is just to easily show / hide the eventbody before it goes through the parsing functions and the services: I expect a customer to uncomment it only when something goes wrong with the webhooks, such as the event name has changed on the sinch platform, or there is an unspecified event that is not taken care of.
I like the idea of a wrapper to enable / disable logs but in this precise case of using a webhook for a single domain, I would want to impact the behaviour for other domains and flood the console

packages/sdk-client/src/utils/authorization.helper.ts
asein-sinch: Fixed, the is now returning a boolean

packages/voice/src/rest/v1/applications/applications-api.ts
asein-sinch: I've added a comment

JPPortier (approved):

PR #12, opened 2024-01-24
Provide options for unexploded parameters

asein-sinch:
The OAS documents are not specifying the arrays of query parameters correctly:
 - in Numbers (List active numbers)
```
        - name: capability
          in: query
          description: Number capabilities to filter by, `SMS` and/or `VOICE`.
          style: form
          explode: true
          schema:
            type: array
            items:
              type: string
```
==> The server accepts `capability=SMS;VOICE` (mind the semicolon)
==> The server rejects `capability=SMS,VOICE` (mind the comma): "message": "parsing field \"capability\": \"SMS,VOICE\" is not a valid value",
==> The server rejects `capability=SMS&capability=VOICE`: "message": "too many values for field \"capability\": SMS, VOICE",
 - in conversation
```
    update_mask.paths:
      description: The set of field mask paths.
      name: update_mask.paths
      in: query
      explode: true
      schema:
        type: array
        items:
          type: string
```
==> The server accepts `update_mask=channel_priority,channel_identities,display_name`

Both specifications are incorrect as it should be `exploded: false` and the first one is even more incorrect as the parameters should be coma-separated.
Reference: https://swagger.io/docs/specification/serialization/

### DIFF
--- a/packages/sdk-client/src/api/api-client.ts
+++ b/packages/sdk-client/src/api/api-client.ts
@@ -75,9 +75,9 @@ export class ApiClient {
       .reduce(
         (acc, name) => {
           const prop = data[name];
-          acc[name] = typeof prop.toJSON === 'function'
-            ? prop.toJSON()
-            : Array.isArray(prop) ? prop.join(',') : prop.toString();
+          acc[name] = typeof prop.toJSON === 'object'
+            ? JSON.stringify(prop.toJSON())
+            : JSON.stringify(prop);
           return acc;
         },
         {} as { [p in keyof T]: string },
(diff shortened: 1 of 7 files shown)


JPPortier:
About comment
> The server accepts capabilities=SMS;VOICE (mind the semicolon)

But `?regionCode=US&type=LOCAL&capabilities=SMS&capabilities=VOICE `
Is accepted by server and conform to https://swagger.io/docs/specification/serialization for `explode=true` and `form=true`
image

packages/sdk-client/src/api/api-client.ts
```diff
@@ -77,7 +77,7 @@ export class ApiClient {
           const prop = data[name];
           acc[name] = typeof prop.toJSON === 'function'
             ? prop.toJSON()
-            : Array.isArray(prop) ? prop.join(';') : prop.toString();
+            : Array.isArray(prop) ? prop.join(',') : prop.toString();
```
JPPortier, line 80: Does this fit if any array item contains a comma character? 
This could have side effect to decode it later. Right ?

asein-sinch:
I've update the description with the right operation that causes some issues. Sorry for the confusion

JPPortier (approved):
Agree about OAS inconsitency.
For numbers and list operation: available vs active
Both are described as `array` / `form` / `explode`
but when requesting with multiple values : 
- **active** response: `INVALID_ARGUMENT: too many values for field "capability": SMS, VOICE `
- **available** response: `200 / OK` and response is filled

JPPortier (approved):

PR #13, opened 2024-01-24
Conversation API support

asein-sinch:
Branch to be merged on `main` when the official documentation will be delivered


packages/conversation/src/rest/v1/app/app-api.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+export interface ListAppsRequestData {
+}
+export interface UpdateAppRequestData {
+  /** The unique ID of the app. You can find this on the [Sinch Dashboard](https://dashboard.sinch.com/convapi/apps). */
+  'app_id': string;
+  /** The updated app. */
+  'appUpdateRequestBody': AppUpdateRequest;
+  /** The set of field mask paths. */
+  'update_mask'?: Array<string>;
```
JPPortier, line 33: What about simplifying SDK usage but computing `update_mask` from provided `appUpdateRequestBody` content ?
e.g.
- appUpdateRequestBody do not contains a `display_name` value (i.e. `undefined` or key is not present): the update_mask will not contains `display_name` keyword to not handle the value from payload
- appUpdateRequestBody  contains a `display_name` value (i.e. `null` or `any string`): the update_mask will contains `display_name` keyword to set the value according to payload

Then the `update_mask`is no longer required on function call and it simplify call and logic onto SDK usage
This could improve DX (see https://github.com/sinch/sinch-sdk-node/pull/15/files#diff-5f095c61b28d52390193b18e6684b856802e801978662161d34257e25e9d0b26R18) and providing added value from SDKs above pure REST calls

Dovchik (commented):
Part one

packages/sdk-core/src/sinch-client.ts
```diff
@@ -3,10 +3,12 @@ import { Sms } from '@sinch/sms';
 import { Verification } from '@sinch/verification';
 import { SinchClientParameters } from '@sinch/sdk-client';
 import { Voice } from '@sinch/voice';
+import { ConversationDomain } from '@sinch/conversation';
```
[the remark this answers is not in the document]
asein-sinch: Because `Conversation` is already the name of an exported model

packages/conversation/src/models/v1/enums.ts
```diff
@@ -0,0 +1 @@
+export {};
```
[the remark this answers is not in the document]
asein-sinch: This file is generated in order to add enums manually when needed

packages/sdk-core/src/sinch-client.ts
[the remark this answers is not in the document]
Dovchik: AH, dunno, the break in consistency in top api level seems hits me harder than on model level
Isn't it better to rename the model to `ConversationModel`?

packages/conversation/src/rest/v1/app/app-api.ts
asein-sinch: I don't concur with the analysis above: the user may want to send some data in the body (for instance copied from a previous `GET` request from the contact or another contact) and update only a subpart of it: this way, by specifying the property in the `update_mask` all the other properties will be ignored.
Suppressing the parameter in the public api and/or filling it automatically may improve DX but it degrades the API flexibility. Hence I'm not in favour of removing it from the request data interface

packages/sdk-core/src/sinch-client.ts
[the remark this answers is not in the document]
asein-sinch: `ConversationModel` would then differ from the online documentation. 
`ConversationDomain` is used only by the SinchClient to initialize the class and should not be used by the end user.
Either way it's not ideal

packages/conversation/src/rest/v1/app/app-api.ts
Dovchik: What about adding override function, one with auto `update_mask`
```
Update(Contact)
```
And second for distinguished users which will override automatically set:
```
Update(Contact, UpdateMask)
```

Dovchik (approved):

packages/conversation/src/rest/v1/app/app-api.ts
Dovchik: > I don't concur with the analysis above: the user may want to send some data in the body (for instance copied from a previous `GET` request from the contact or another contact) and update only a subpart of it: this way, by specifying the property in the `update_mask` all the other properties will be ignored. Suppressing the parameter in the public api and/or filling it automatically may improve DX but it degrades the API flexibility. Hence I'm not in favour of removing it from the request data interface

To add: if you won't touch other properties in this case they will be the same as before anyway so you will still achieve desired result

Dovchik:
```
var contact = Get() { 'display_name': 'a', 'channel': 'b' }
contact.channel = 'c'
Update(contact) { 'display_name': 'a', 'channel': 'c' }
```
The the same result above will be when you set `update_mask` to `channel`.

asein-sinch: I'm sorry but I still don't understand your point: in which use case do you want to force a user to use an parameter defined as optional by the API?

Dovchik: ah, right, it's optional, then doesn't matter

packages/sdk-core/src/sinch-client.ts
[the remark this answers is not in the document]
asein-sinch: Fixed

PR #16, opened 2024-01-30
DEVEXP-288: Support Callback Events in Conversation API

asein-sinch:


packages/conversation/src/models/v1/channel-event/channel-event.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export interface ChannelEventNotification {
+
+  /** @see ConversationChannel */
+  channel?: ConversationChannel;
+  /** The type of event being reported. */
+  event_type?: string;
+  /** An object containing additional information regarding the event. The contents of the object depend on the channel and the event_type. */
+  additional_data?: object;
```
JPPortier, line 33: Do we have more information about `object` from OAS ?
(just interested because for Java SDK it could be an issue)

packages/conversation/src/models/v1/message-inbound-event-item/message-inbound-event-item.ts
```diff
@@ -0,0 +1,29 @@
...
+  /** The ID of the contact. Will be empty if processing_mode is DISPATCH. */
+  contact_id?: string;
+  /** Usually, metadata specific to the underlying channel is provided in this field. Refer to the individual channels\' documentation for more information (for example, SMS delivery receipts). Note that, for Choice message responses, this field is populated with the value of the message_metadata field of the corresponding Send message request. */
+  metadata?: string;
+  /** Timestamp marking when the channel callback was received by the Conversation API. */
+  accept_time?: string;
+  /** The sender ID to which the contact sent the message, if applicable. For example, originator msisdn/short code for SMS and MMS. */
+  sender_id?: string;
+  /** Whether or not Conversation API should store contacts and conversations for the app. For more information, see [Processing Modes](../../../../../conversation/processing-modes/). */
+  processing_mode?: string;
```
JPPortier, line 25: May be an enum type if list of expected values is available

asein-sinch: Ah yes, I missed this one

packages/conversation/src/models/v1/channel-event/channel-event.ts
asein-sinch: Unfortunately not: we have an object with 3 properties and the 3rd one depends on the 2 others. We have a list of `channel`, but we don't have a list of `event_type`.
So for the moment, the `additionnal_data` is treated as a data map.
Here is the example from the spec:
```json
{
  "channel": "WHATSAPP",
  "event_type": "WHATS_APP_QUALITY_RATING_CHANGED",
  "additional_data": {
    "quality_rating": "GREEN"
  }
}
```

packages/conversation/src/models/v1/conversation-delete-event/conversation-delete-event.ts
```diff
@@ -0,0 +1,33 @@
...
+  /** The project ID of the app which has subscribed for the callback. */
+  project_id?: string;
+  /** Context-dependent metadata. Refer to specific callback\'s documentation for exact information provided. */
+  message_metadata?: string;
+  /** The value provided in field correlation_id of a send message request. */
+  correlation_id?: string;
+  /** @see ConversationNotification */
+  conversation_delete_notification?: ConversationNotification;
+
+  trigger: 'CONVERSATION_DELETE';
```
Dovchik, line 23: Maybe makes sense to make enum for types of triggers to have a unified list in one place

examples/webhooks/src/controller/app.controller.ts
```diff
@@ -27,6 +33,26 @@ export class AppController {
     private readonly verificationService: VerificationService,
     private readonly voiceService: VoiceService) {}
 
+  @Post('/conversation')
+  public conversation(@Req() request: Request, @Res() res: Response) {
+    // console.log(request.headers);
+    // console.log(JSON.stringify(request.body, null, 2));
+    // console.log(request['rawBody']);
```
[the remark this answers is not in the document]
asein-sinch: It's there on purpose for debugging needs

packages/conversation/src/models/v1/conversation-delete-event/conversation-delete-event.ts
asein-sinch: It's the type `WebhookTrigger` that defines all the possible values

examples/webhooks/src/services/conversation.service.ts
```diff
@@ -1,9 +1,212 @@
...
+
+  private initClient = () => {
+    const keyId = process.env.SINCH_KEY_ID || '';
+    const keySecret = process.env.SINCH_KEY_SECRET || '';
+    const projectId = process.env.SINCH_PROJECT_ID || '';
+    return new SinchClient({ projectId, keyId, keySecret });
+  };
+
+  private buildTextMessage(contactMessage: ContactMessage) {
+    if ('text_message' in contactMessage) {
```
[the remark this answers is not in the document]
asein-sinch: No, this is a oneOf with no discriminator

JPPortier (approved):

packages/conversation/src/models/v1/conversation-delete-event/conversation-delete-event.ts
Dovchik: can you do smth like that: `trigger: WebhookTrigger.CONVERSATION_DELETE` then?

asein-sinch: No, this is not an enum but a type. I've created an new interface to ensure the trigger value matches the type

PR #17, opened 2024-01-31
Support multiple regions for the conversation API

asein-sinch:


packages/conversation/src/rest/v1/conversation-domain-api.ts
```diff
@@ -59,9 +61,13 @@ export class ConversationDomainApi implements Api {
    */
   public getSinchClient(): ApiClient {
     if (!this.client) {
+      const region = this.sinchClientParameters.region || Region.UNITED_STATES;
+      if(!Object.values(ConversationRegion).includes((region as unknown) as ConversationRegion)) {
+        throw new Error(`The region '${region}' is not supported for the Conversation API`);
```
JPPortier, line 66: What about only logging an error/warning?
Imagine a new region is supported next months but SDK is not yet ready nor released. 
An end user wish to test the new region (it was one of it request to support) but he won't be able to. Worst if the region is in beta test it mean we have to release a new version to test a alpha/beta feature just due to a config level setting.

packages/conversation/src/rest/v1/conversation-domain-api.ts
```diff
@@ -59,9 +61,13 @@ export class ConversationDomainApi implements Api {
...
+      }
       const apiClientOptions = this.buildApiClientOptions(this.sinchClientParameters);
       this.client = new ApiFetchClient(apiClientOptions);
-      this.client.apiClientOptions.basePath = 'https://us.conversation.api.sinch.com';
+      this.client.apiClientOptions.basePath = `https://${region}.conversation.api.sinch.com`;
```
JPPortier, line 70: Then currently the region is only to set the URL, but for internal or external validation purpose are we able to set a specific URL value ?
e.g. How to use `https:/foo.com` in place of `https://${region}.conversation.api.sinch.com`
- the value is set by end user
- there is not format to apply in regard of region value

asein-sinch: The method `setBasePath()` is there for that :) 
I've added a test to illustrate its usage.

packages/conversation/src/rest/v1/conversation-domain-api.ts
asein-sinch: Thanks for your comment. I've updated the code and the test accordingly

packages/conversation/tests/rest/v1/conversation-domain-api.test.ts
```diff
@@ -26,11 +26,23 @@ describe('Conversation API', () => {
     expect(conversationApi.client?.apiClientOptions.basePath).toBe('https://eu.conversation.api.sinch.com');
   });
 
-  it('should throw an error when using an unsupported region', () => {
+  it('should throw an error when using an unsupported region', async () => {
```
JPPortier, line 29: NIT: the title test is no longer reflecting the test

packages/conversation/src/rest/v1/conversation-domain-api.ts
JPPortier: ok. overriding the value have to be done after initialization
This emphasizes that the exception must not be raised during initialization 👍

JPPortier (approved):

packages/conversation/tests/rest/v1/conversation-domain-api.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should change the URL when specifying a different region', () => {
+    params.region = Region.EUROPE;
+    conversationApi = new ConversationApi(params);
+    conversationApi.getSinchClient();
+    expect(conversationApi.client?.apiClientOptions.basePath).toBe('https://eu.conversation.api.sinch.com');
+  });
+
+  it('should log a warning when using an unsupported region', async () => {
+    params.region = Region.CANADA;
```
Dovchik, line 30: Should it be possible at all? I mean why Regions is generic - when you can provide conversationRegion param - I even see below the dedicated region enum for Conversation

asein-sinch: Javascript allows it all

PR #20, opened 2024-02-08
DEVEXP-290: Support Templates API v1 and v2

asein-sinch:
Support for the TemplatesAPI v1 and v2: 
 - both services are registered on the conversation controller
 - introduction of helper methods to simplify the template content creation


JPPortier (approved):
Approval related to functionality is available for template V1 & V2.
As open point requiring more discussion outside of this PR only: curious if we can move to an abstraction for users about "just" supporting "template" and do not show them V1 vs V2 transition/complexity and or even provide a smooth transition from V1 to V2 for existing implementation: let discuss this outside of PR review

PR #24, opened 2024-02-21
DEVEXP-269 (Part 1/2): Refactor Voice models

asein-sinch:
This PR is mainly about some refactoring among generated models. Nothing major here.

Only notable change: `voice.applications.getNumbers` becomes `voice.applications.listNumbers` to stay consistent with our naming => **To be updated in other SDKs**


packages/voice/src/models/v1/call-object/call-object.ts
```diff
@@ -0,0 +1,31 @@
+import { ReasonEnum, ResultEnum } from '../enums';
+import { VoicePrice } from '../voice-price';
+import { Participant } from '../participant';
+
+export interface CallObject {
```
JPPortier, line 5: It took me some time to identify what mean "CallObject" and it usage.
Finally have to ensure I was right by looking at the OAS file details to identify it is the `getCallResponseObj` schema.
So finally, may be it could mean the name should convey more information for auto-documentation  purpose and readability for maintainers

JPPortier (commented):
Not reviewed all generated files, but it could be interesting to identify from oas generator tool repo itself what cause these refactoring onto existing sources
So: we won't have to review generated sources but the generator itself and it enhancements

packages/voice/src/models/v1/call-object/call-object.ts
asein-sinch: Naming is a complicated task :)
The main object here is a 'Call' which is also the name of the service 'Calls' and also the verb 'call'. This is why I've added the suffix 'Object' to be clear about which kind of call it was and to avoid a naming conflict.
Then, the operationId is `Calling_GetCallResult` but here "Result" doesn't mean "Response" but rather "Status" which confused me too. The response object defined in the specification is `getCallResponseObj` which is "the object that represents the response of a getCall (status) query". So I thought to name it "CallResponse" but then it would be confusing with the actions of the "Call" service.
Long story short: what do you think about `CallInformation`?

JPPortier: Yes; and it is not so far from documentation related to operation: `Get information about a call`

asein-sinch: Fixed. In the end it's `GetCallInformation` as it is the result of a `GET` call, such as "get callbacks" for instance

JPPortier (approved):

PR #25, opened 2024-02-21
DEVEXP-269 (Part 2/2): Add helpers for building SVAML actions and instructions

asein-sinch:
Please check in particular the 2 approaches to build a SVAML object in the `helper.ts` file:
 - builder pattern
 - rest parameters (variadic number of `instruction`)


packages/voice/src/models/v1/helper.ts
```diff
@@ -0,0 +1,239 @@
...
+
+export class IceSvamletBuilder extends SvamletBuilder<IceSvamlAction, IceSvamlInstruction, IceResponse> {}
+
+export class PieSvamletBuilder extends SvamletBuilder<PieSvamlAction, PieSvamlInstruction, PieResponse> {}
+
+export const customCalloutHelper = {
+  formatIceResponse: (action?: IceSvamlAction, ...instructions: IceSvamlInstruction[]): string => {
+    const iceResponse: IceResponse = {
+      action,
+      instructions: instructions.length > 0 ? instructions : undefined,
```
JPPortier, line 69: `SvamletBuilder` constructor do not initialize with a default value the `instructions` field
So, could we imagine having an exception thrown here ?

packages/voice/src/models/v1/helper.ts
```diff
@@ -0,0 +1,239 @@
...
+  buildSay: (text: string, locale?: TtsVoice): SvamlInstructionSay => {
+    return {
+      name: 'say',
+      text,
+      locale,
+    };
+  },
+  buildSendDtmf: (dtmfValue: string): SvamlInstructionSendDtmf => {
+    const dtmfRegex = /^[0-9#w]+$/;
+    if(!dtmfRegex.test(dtmfValue)) {
```
JPPortier, line 161: Java SDK is having a relax mode regarding validation and up to user to activate it (strict mode is the default one)
So still in place with log but won't throw exception
Idea is, in case of user do not have possibility to fix the DTMF source value, to avoid a blocking situation at user level because of SDK just refuse to perform the request
Java SDK implementation: https://github.com/sinch/sinch-sdk-java/blob/main/client/src/main/com/sinch/sdk/models/DualToneMultiFrequency.java

packages/voice/src/models/v1/music-on-hold/music-on-hold.ts
```diff
@@ -0,0 +1 @@
+export type MusicOnHold = 'ring' | 'music1' | 'music2' | 'music3';
```
JPPortier, line 1: Why sometime a type is defined for supported enum values like here with  `MusicOnHold` or `TtsVoice` and sometime the authorized values are directly set at "usage" level  (see the PR https://github.com/sinch/sinch-sdk-node/pull/24/files#diff-310899169f21ebf8d587f23f4b09b4e15353f657ff6b7dd7a705807229ab0e73R8) ?

examples/webhooks/src/services/voice.service.ts
```diff
@@ -81,36 +78,28 @@ export class VoiceService {
...
-      instructions: [
-        {
-          name: 'say',
-          text: 'Thanks for your input. The call will now end.'
-        }
-      ]
-    }
+    const pieResponse = new PieSvamletBuilder()
+      .setAction(pieActionHelper.hangup())
+      .addInstruction(pieInstructionHelper.say('Thanks for your input. The call will now end.'))
```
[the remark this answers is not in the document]
asein-sinch: Indeed, this will not work and the call will be disconnected immediately :(

packages/voice/src/models/v1/helper.ts
asein-sinch: No, instructions are optional. In this approach with rest parameters, if no instructions are provided, the serialization will produce an `instructions` property with an empty array, which is not what we want. 
In the `SvamletBuilder`, if there is an empty array, this is because the user specifically added this empty array, so I let it be serialized (honestly, I don't know the impact of having an empty array versus no property at all)

packages/voice/src/models/v1/helper.ts
asein-sinch: I'm wondering if we should block the DTMF action at SDK level (and not perform the intended action), or let the action to occur on the backend (and fail), which would have the exact same experience to the end user on the line.
Logging the error is the most important thing to do to help the developer understand what may have gone wrong.
BTW, there is a typo in your error message on line 35: "Invalid DTFM format" (meta review!!😁)

packages/voice/src/models/v1/music-on-hold/music-on-hold.ts
asein-sinch: It depends 😅
Generally, when the type is super simple (one or two values), I define it inline so it's still readable (this is the case for all the actions and instructions names for instance)
When the type is used several times, I prefer to create a dedicated one, such as `MusicOnHold` which is used a couple of times
And when the type is super complex (like the supported locales or the conversation triggers), it deserves it's own type.
But writing these lines, I realized that I'm not consistent: sometimes I create a dedicated file, sometimes I group the enum types in the `enum.ts` file. I'll have to refactor this!

packages/voice/src/models/v1/helper.ts
JPPortier: As far as I know, currently, server could silently ignore inconsistent value
(thank you for the Java review 😄 )

examples/webhooks/src/services/voice.service.ts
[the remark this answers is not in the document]
asein-sinch: fixed

packages/voice/src/models/v1/music-on-hold/music-on-hold.ts
asein-sinch: Fixed

JPPortier (approved):

PR #27, opened 2024-02-21
DEVEXP-192: Verification flow example

asein-sinch:
Example of a an interactive verification flow on all methods


JPPortier (approved):
Good job.
We'll have to extend the Java interactive workflow to offer more than just SMS verification

PR #32, opened 2024-02-28
DEVEXP-316: Add support for Faxes

asein-sinch:
Add support for faxes and it needed some changes in the API client to support the multipart/formdata


packages/fax/src/rest/v3/faxes/faxes-api.ts
```diff
@@ -0,0 +1,285 @@
...
+    // the request will be sent as multipart/formdata
+    const isUsingJson = (data['sendFaxRequestBody'] as FaxRequestJson).files !== undefined;
+    if (isUsingJson) {
+      headers['Content-Type'] = 'application/json';
+      body = JSON.stringify(data['sendFaxRequestBody']);
+    } else {
+      const formParams:any = {};
+      const requestData = data.sendFaxRequestBody as FaxRequestFormData;
+      if( requestData.to ) { formParams['to'] = requestData.to; }
+      if( requestData.file ) { formParams['file'] = requestData.file; }
```
spacedsweden, line 244: this faxrequest body contains a ton of readonly fields you can not set, price, barcodes etc please refer to https://developers.sinch.com/docs/fax/api-reference/fax/tag/Faxes/#tag/Faxes/operation/sendFax for valid in parameters.

**Question** : where are the files proccessed for multpard and applicaiton json, (base64 for json.
**Question 2**: can you show me an example on how its sent? and could we have a few money maker methods for docs and the normal use

example 
sendFax(to,from,file)
sendFax(to,from,contenturl)

In the sdk im not sure we want to expose that a customer can choose json or form data, i think it would be better that we always post multipart formdata its more effective and hide that fact for teh sdk user. no need to complicate it

asein-sinch: Then it would be good to not define those fields as part of the request body in the documentation: at the moment it mentions:
 - to (required)
 - file (optional)
 - from (optional)
 - contentUrl (optional)
 - headerText (optional) - default: ""
 - headerPageNumbers (optional) - default: true
 - headerTimeZone (optional) - default: "America/New_York"
 - retryDelaySeconds (optional) - default: 60
 - labels (optional)
 - callbackUrl (optional)
 - callbackContentType (optional) - default "multipart/form-data"
 - imageConversionMethod (optional) - default "HALFTONE"
 - serviceId (optional)
 - maxRetries (optional)
 The SDK is not meant to downgrade the possibilities offered by the API so all these fields are exposed. If they are not meant to be exposed to the customer, then please update the specification accordingly, I'll then happily update the request interface.

asein-sinch: > Question : where are the files proccessed for multpard and applicaiton json, (base64 for json.

I'm not sure to understand your question, can you please clarify?

spacedsweden: > Then it would be good to not define those fields as part of the request body in the documentation: at the moment it mentions:
> 
> * to (required)
> * file (optional)
> * from (optional)
> * contentUrl (optional)
> * headerText (optional) - default: ""
> * headerPageNumbers (optional) - default: true
> * headerTimeZone (optional) - default: "America/New_York"
> * retryDelaySeconds (optional) - default: 60
> * labels (optional)
> * callbackUrl (optional)
> * callbackContentType (optional) - default "multipart/form-data"
> * imageConversionMethod (optional) - default "HALFTONE"
> * serviceId (optional)
> * maxRetries (optional)
>   The SDK is not meant to downgrade the possibilities offered by the API so all these fields are exposed. If they are not meant to be exposed to the customer, then please update the specification accordingly, I'll then happily update the request interface.

Compare that to the current request in node and you will see for example price is a request parameter and it should not be and it’s not in the docs. It looks the request is infact the complete faxobject

spacedsweden: > > In the sdk im not sure we want to expose that a customer can choose json or form data,
> 
> Again, the SDK doesn't aim to restrict the API possibilities. According to what I see in the documentation, the only contentType default is the one for the callback. If the Fax API PM want to promote formdata as a default request, this needs to be documented and communicated to us so that we act accordingly in the SDK.

I can confirm as the api manager that I would like to promote formencided over json when sending faxes just because it’s more efficient

spacedsweden: > > could we have a few money maker methods for docs and the normal use
> 
> This has to be discussed with the SDK team to decide how to present such method, versus the complete ones

Ok I see similar need in sms tbh especially in documentation we need to beat competition in lines of code and clarity

asein-sinch: > Compare that to the current request in node and you will see for example price is a request parameter and it should not be and it’s not in the docs. It looks the request is infact the complete faxobject

I see it now, I've removed them. Thanks

asein-sinch: > I can confirm as the api manager that I would like to promote formencided over json when sending faxes just because it’s more efficient

> Ok I see similar need in sms tbh especially in documentation we need to beat competition in lines of code and clarity

Interesting topics! 👍 For both these comments, I'll let you schedule a meeting with our manager to discuss further and plan accordingly

spacedsweden: > > Compare that to the current request in node and you will see for example price is a request parameter and it should not be and it’s not in the docs. It looks the request is infact the complete faxobject
> 
> I see it now, I've removed them. Thanks

you removed all, like pagecount, status etc?

asein-sinch: I've updated the fax request model to match strictly with the online documentation: the diff can be seen in the commit https://github.com/sinch/sinch-sdk-node/pull/32/commits/6691399e2229a9a1a564e989fbfa68ad58a6568a

examples/simple-examples/src/fax/faxes/list.ts
```diff
@@ -0,0 +1,67 @@
...
+  const requestData: ListFaxesRequestData = {
+    pageSize: 2,
+  };
+
+  const sinchClient = initClient();
+
+  // ----------------------------------------------
+  // Method 1: Fetch the data page by page manually
+  // ----------------------------------------------
+  let response= await sinchClient.fax.faxes.list(requestData);
```
650elx, line 29:
```suggestion
  let response = await sinchClient.fax.faxes.list(requestData);
```

asein-sinch: Fixed

examples/webhooks/src/services/voice.service.ts
650elx, line 1: Why there are Voice related files and changes within this PR?
I don't see any information within DEVEXP-316 regarding Voice API.
Please move those changes to separate PR/ticket (this PR is already bloated).

650elx (changes requested):

packages/fax/src/models/v3/fax/fax.ts
```diff
@@ -0,0 +1,62 @@
+import { BarCode } from '../bar-code';
+import { FaxContentUrl } from '../fax-content-url';
+import { FaxMoney } from '../fax-money';
+import { ErrorType, FaxDirection, FaxStatus, ImageConversionMethod, WebhookContentType } from '../enums';
+
+export interface Fax {
+
+  /** The id of a fax */
+  id?: string;
```
650elx, line 9: Is it necessary to have all fields within this interface marked as optional?
`id` seems to be like not an optional thing.

examples/webhooks/tsconfig.json
```diff
@@ -12,10 +12,11 @@
     "baseUrl": "./",
     "incremental": true,
     "skipLibCheck": true,
-    "strictNullChecks": false,
+    "strictNullChecks": true,
```
650elx, line 15: Nice 👌

asein-sinch: Unfortunately, I'll removed it from this PR due to the requested changes above

packages/fax/src/models/v3/fax/fax.ts
asein-sinch: This is what the specifications says

examples/webhooks/src/services/voice.service.ts
asein-sinch: Ticket created for it: https://tickets.sinch.com/browse/DEVEXP-343
WIP to remove it from this PR

packages/fax/src/rest/v3/callbacks/callbacks-webhook.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export class FaxCallbackWebhooks implements CallbackProcessor<FaxWebhookEventParsed> {
+
+  /**
+   * Reviver for a Fax Event
+   * This method ensures the object can be treated as a Fax Event and should be called before any action is taken to manipulate the object.
+   * @param {any} eventBody - the event body or form containing the Fax event notification.
+   * @return {FaxWebhookEventParsed} - The parsed Fax event object
+   */
+  public parseEvent(eventBody: any): FaxWebhookEventParsed {
```
650elx, line 15: Is `any` necessary here?
Maybe it's worth giving `eventBody` a type?

examples/webhooks/src/services/voice.service.ts
asein-sinch: Done

packages/fax/src/rest/v3/callbacks/callbacks-webhook.ts
asein-sinch: This is precisely the purpose of this method: an unknown object is received as an event. This method checks its properties to determine it can be processed as a Fax Event.

packages/fax/src/rest/v3/faxes/faxes-api.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  /**
+   * Get fax
+   * Get fax information using the ID number of the fax.
+   * @param { GetFaxRequestData } data - The data to provide to the API call.
+   */
+  public async get(data: GetFaxRequestData): Promise<Fax> {
+    this.client = this.getSinchClient();
+    const getParams = this.client.extractQueryParams<GetFaxRequestData>(data, [] as never[]);
+    const headers: { [key: string]: string | undefined } = {
```
650elx, line 130: Correct me if I'm wrong but it looks like all methods within this class have the same `headers` var defined?
Maybe a class variable would be more concise?

asein-sinch: This is because the specification defines only endpoint with these headers. This code is generated from the specification and it's not a problem to have these variables repeated. Conciseness is not sought here: if the specification changes, it will be over-complicated in the mustache templates to revert the assignment of a variable

650elx: Code duplication is simply eye catching :)  It's up to you to decide then.

650elx (approved):

PR #34, opened 2024-02-29
DEVEXP-320: Add support for Emails

asein-sinch:


examples/simple-examples/src/fax/emails/list.ts
```diff
@@ -0,0 +1,67 @@
...
+  const requestData: ListEmailsForProjectRequestData = {
+    pageSize: 2,
+  };
+
+  const sinchClient = initClient();
+
+  // ----------------------------------------------
+  // Method 1: Fetch the data page by page manually
+  // ----------------------------------------------
+  let response= await sinchClient.fax.emails.list(requestData);
```
650elx, line 29:
```suggestion
  let response = await sinchClient.fax.emails.list(requestData);
```

examples/simple-examples/src/fax/emails/listNumbers.ts
```diff
@@ -0,0 +1,74 @@
...
+    email,
+    pageSize: 2,
+  };
+
+  const sinchClient = initClient();
+
+  // ----------------------------------------------
+  // Method 1: Fetch the data page by page manually
+  // ----------------------------------------------
+  let response= await sinchClient.fax.emails.listNumbers(requestData);
```
650elx, line 36:
```suggestion
  let response = await sinchClient.fax.emails.listNumbers(requestData);
```

examples/simple-examples/src/fax/services/listEmailsForNumber.ts
```diff
@@ -0,0 +1,65 @@
...
+    serviceId,
+    pageSize: 2,
+  };
+
+  const sinchClient = initClient();
+
+  // ----------------------------------------------
+  // Method 1: Fetch the data page by page manually
+  // ----------------------------------------------
+  let response= await sinchClient.fax.services.listEmailsForNumber(requestData);
```
650elx, line 29:
```suggestion
  let response = await sinchClient.fax.services.listEmailsForNumber(requestData);
```

packages/fax/src/rest/v3/emails/emails-api.ts
```diff
@@ -0,0 +1,224 @@
...
+   * Add an email to be used for sending and receiving faxes.
+   * @param { AddEmailToNumbersRequestData } data - The data to provide to the API call.
+   */
+  public async addToNumbers(data: AddEmailToNumbersRequestData): Promise<Email> {
+    this.client = this.getSinchClient();
+    const getParams = this.client.extractQueryParams<AddEmailToNumbersRequestData>(data, [] as never[]);
+    const headers: { [key: string]: string | undefined } = {
+      'Content-Type': 'application/json',
+      'Accept': 'application/json',
+    };
```
650elx, line 64: Those headers seem to be the same for all `EmailsApi` operation. Maybe it will be better to not duplicate the code and use class variable for that?

examples/simple-examples/src/fax/emails/list.ts
asein-sinch: Fixed

examples/simple-examples/src/fax/emails/listNumbers.ts
asein-sinch: Fixed

packages/fax/src/rest/v3/emails/emails-api.ts
asein-sinch: I don't think it would be better to factorize in the code what can be (and actually is) defined individually at operation level in the specification. If the specification is updated later on and changes a single operation definition, the impacts on this generated code will be much bigger to apprehend than a direct regeneration.
What is bothering you exactly in this file? Is it the repetition of the same string? Is it the object definition itself?

examples/webhooks/src/services/voice.service.ts
650elx, line 1: Looks like similar situation as in #32

packages/fax/src/rest/v3/emails/emails-api.ts
650elx: yeah; just the repetition thing. Personally I like more concise source files.

650elx (approved):

examples/webhooks/src/services/voice.service.ts
asein-sinch: Removed. See PR#36

PR #48, opened 2024-03-29
Refactor webhooks examples

asein-sinch:


examples/webhooks/src/controller/app.controller.ts
```diff
@@ -9,12 +9,12 @@ import {
   VerificationCallbackWebhooks,
   VoiceCallbackWebhooks,
 } from '@sinch/sdk-core';
-import { NumbersService } from '../services/numbers.service';
-import { SmsService } from '../services/sms.service';
-import { VerificationService } from '../services/verification.service';
-import { VoiceService } from '../services/voice.service';
-import { ConversationService } from '../services/conversation.service';
-import { FaxService } from '../services/fax.service';
+import { NumbersEventService } from '../services/numbers-event.service';
```
JPPortier, line 12: Like the consistent terminology "EventService" claiming what the service is related to.
We could imagine consistent usage across SDKs and why not endpoint (/conversation, /sms, ...) 
Should be very helpful from SDK's documentation too to move from a REST API documentation to an SDK documentation

JPPortier (approved):

PR #51, opened 2024-04-04
DEVEXP-379: Support hostname override in SinchClient parameters

asein-sinch:
Important files to review are the following ones (others are refactoring and linting):

- *-domain-api.ts
- *-domain-api.test.ts

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,6 +32,7 @@
       "@typescript-eslint/no-var-requires": "off",
       "semi": "warn",
       "comma-dangle": "warn",
+      "quotes": ["warn", "single"],
       "jest-extended/prefer-to-be-true": "warn",
       "jest-extended/prefer-to-be-false": "error",
       "jest-formatting/padding-around-describe-blocks": 2,
--- a/packages/conversation/src/rest/v1/conversation-domain-api.ts
+++ b/packages/conversation/src/rest/v1/conversation-domain-api.ts
@@ -104,7 +104,7 @@ export class ConversationDomainApi implements Api {
     switch (this.apiName) {
     case 'TemplatesV1Api':
     case 'TemplatesV2Api':
-      return this.sinchClientParameters.templatesHostname ?? `https://${region}.template.api.sinch.com`;
+      return this.sinchClientParameters.conversationTemplatesHostname ?? `https://${region}.template.api.sinch.com`;
     default:
       return this.sinchClientParameters.conversationHostname ?? `https://${region}.conversation.api.sinch.com`;
     }
--- a/packages/conversation/src/rest/v1/conversation-domain-api.ts
+++ b/packages/conversation/src/rest/v1/conversation-domain-api.ts
@@ -41,7 +41,7 @@ export class ConversationDomainApi implements Api {
   public setRegion(region: Region) {
     this.sinchClientParameters.region = region;
     if (this.client) {
-      this.client.apiClientOptions.hostname = this.buildBasePath(region);
+      this.client.apiClientOptions.hostname = this.buildHostname(region);
     }
   }
 
@@ -83,8 +83,7 @@ export class ConversationDomainApi implements Api {
       }
       const apiClientOptions = this.buildApiClientOptions(this.sinchClientParameters);
       this.client = new ApiFetchClient(apiClientOptions);
-      this.client.apiClientOptions.hostname = this.sinchClientParameters.conversationHostname
-        ?? this.buildBasePath(region);
+      this.client.apiClientOptions.hostname = this.buildHostname(region);
     }
     return this.client;
   }
@@ -101,13 +100,13 @@ export class ConversationDomainApi implements Api {
     };
   }
 
-  private buildBasePath(region: Region) {
+  private buildHostname(region: Region) {
     switch (this.apiName) {
     case 'TemplatesV1Api':
     case 'TemplatesV2Api':
-      return `https://${region}.template.api.sinch.com`;
+      return this.sinchClientParameters.templatesHostname ?? `https://${region}.template.api.sinch.com`;
     default:
-      return `https://${region}.conversation.api.sinch.com`;
+      return this.sinchClientParameters.conversationHostname ?? `https://${region}.conversation.api.sinch.com`;
     }
   }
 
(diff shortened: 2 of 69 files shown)


packages/conversation/src/rest/v1/conversation-domain-api.ts
```diff
@@ -83,7 +83,8 @@ export class ConversationDomainApi implements Api {
       }
       const apiClientOptions = this.buildApiClientOptions(this.sinchClientParameters);
       this.client = new ApiFetchClient(apiClientOptions);
-      this.client.apiClientOptions.basePath = this.buildBasePath(region);
+      this.client.apiClientOptions.hostname = this.sinchClientParameters.conversationHostname
```
JPPortier, line 86: This logic lead to end user cannot set a domain to be used dynamically with a region value right?
I mean; setting a host-name to `https://${region}.my-own-domain.foo` then do not change the region handling from his code (say region was set to `US` or not set to use the default one) to use `https://US.my-own-domain.foo`. 
Only way to have this is to set variable `this.sinchClientParameters.conversationHostname` to `https://US.my-own-domain.foo`

Could require a note for SDK's user (_region is not used when hostname is_)

JPPortier, line 86: Another point: may be I missed something from logic but seems you cannot set 2 different hostname to be used for the 2 supported endpoint: `conversation` and `template` endpoints.
You can only override one at same time ?

packages/sms/src/rest/v1/sms-domain-api.ts
```diff
@@ -67,9 +67,8 @@ export class SmsDomainApi implements Api {
     if (!this.client) {
       const region = this.sinchClientParameters.region || Region.UNITED_STATES;
       const apiClientOptions = this.buildApiClientOptions(this.sinchClientParameters, region);
-      const basePath = `https://${apiClientOptions.useServicePlanId?'':'zt.'}${region}.sms.api.sinch.com`;
       this.client = new ApiFetchClient(apiClientOptions);
-      this.client.apiClientOptions.basePath = basePath;
+      this.client.apiClientOptions.hostname = this.sinchClientParameters.smsHostname ?? `https://${apiClientOptions.useServicePlanId?'':'zt.'}${region}.sms.api.sinch.com`;
```
JPPortier, line 71: Same remark as for conversation: design do not enable to set a dynamic domain and use SDK logic to handle regions
Could require a note for SDK's user (_region is not used when hostname is_)

packages/voice/src/rest/v1/voice-domain-api.ts
```diff
@@ -87,15 +87,15 @@ export class VoiceDomainApi implements Api {
       };
       this.client = new ApiFetchClient(apiClientOptions);
       const region: VoiceRegion = this.sinchClientParameters.voiceRegion || VoiceRegion.DEFAULT;
-      this.client.apiClientOptions.basePath = this.buildBasePath(region);
+      this.client.apiClientOptions.hostname = this.sinchClientParameters.voiceHostname ?? this.buildBasePath(region);
```
JPPortier, line 90: Same as for Conversation and Sms: Could require a note for SDK's user (_region is not used when hostname is_)
And same question as for Conversation:  is it possible to override the 2 supported endpoints: the general  one for `Conversation` and the specific one for `Applications` endpoints ?

packages/voice/src/rest/v1/voice-service.ts
```diff
@@ -18,15 +18,15 @@ export class VoiceService {
...
-  public setBasePath(basePath: string) {
-    this.applications.setBasePath(basePath);
-    this.conferences.setBasePath(basePath);
-    this.calls.setBasePath(basePath);
-    this.callouts.setBasePath(basePath);
+  public setHostname(hostname: string) {
+    this.applications.setHostname(hostname);
+    this.conferences.setHostname(hostname);
+    this.calls.setHostname(hostname);
+    this.callouts.setHostname(hostname);
```
JPPortier, line 29: _See comment about possibility to override with different values hostname related to applications and the other ones_

.eslintrc
```diff
@@ -32,6 +32,7 @@
       "@typescript-eslint/no-var-requires": "off",
       "semi": "warn",
       "comma-dangle": "warn",
+      "quotes": ["warn", "single"],
```
JPPortier, line 35: ❤️ 
Better consistency across sources for `"`, `'` and `` ` `` usage

packages/conversation/src/rest/v1/conversation-domain-api.ts
asein-sinch: I've considered that the user would override the hostname to replace it in full, the use case being to use a mock server or a test system, for which I don't think regions would be put in place.
I'll make sure to add it to the documentation before the GA

asein-sinch: Indeed, this is a miss from my end. I'll do it for voice application management too

JPPortier: Yes, use case to override hostname seems to not require handling dedicated region 👍

packages/voice/src/rest/v1/voice-domain-api.ts
asein-sinch: Hehe, I've replied just above already 😁

packages/conversation/src/rest/v1/conversation-domain-api.ts
asein-sinch: Fixed 🙂

packages/voice/src/rest/v1/voice-domain-api.ts
asein-sinch: Fixed 🙂

packages/sdk-client/src/domain/domain-interface.ts
```diff
@@ -38,11 +38,13 @@ export interface ApplicationCredentials {
 export interface ApiHostname {
   authHostname?: string;
   conversationHostname?: string;
+  templatesHostname?: string;
```
JPPortier, line 41: `conversationTemplatesHostName` could have more sense :
- help to identify it is related to conversation domain 
- avoid naming clash and confusion in case other kind of templates will have to be managed in the future for a different domain

(same "good" logic as when `voiceApplicationManagementHostname` was named)

JPPortier (approved):
👍

PR #52, opened 2024-04-05
DEVEXP-380: Support plugins addition in SinchClient parameters

asein-sinch:
- Add `requestPlugins` and `responsePlugins` in the SinchClientParameters
- Refactor the methods to build the ApiClientOptions to factorize the plugins addition
- Add corresponding tests
- Take the opportunity to break the circular dependencies in @sinch/sdk-client package

### DIFF
--- a/packages/sdk-client/src/api/api-client-options-helper.ts
+++ b/packages/sdk-client/src/api/api-client-options-helper.ts
@@ -1,37 +1,87 @@
-// import { ApiTokenRequest, Oauth2TokenRequest, SigningRequest, XTimestampRequest } from '../plugins';
-// // import { ApiClientOptions } from './api-client-options';
-//
-// export const buildApiClientOptionsForProjectId = (
-//   projectId: string,
-//   keyId: string,
-//   keySecret: string,
-// ) => {
-//   return {
-//     projectId,
-//     requestPlugins: [new Oauth2TokenRequest(keyId, keySecret)],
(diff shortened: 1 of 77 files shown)


packages/sdk-client/src/api/api-client-options-helper.ts
```diff
@@ -1,37 +1,87 @@
...
+  return apiClientOptions;
+};
+
+export const buildFlexibleOAuth2OrApiTokenApiClientOptions = (
+  params: SinchClientParameters, region: Region, apiName: string,
+): ApiClientOptions => {
+  let apiClientOptions: ApiClientOptions | undefined;
+  // Check the region: if US or EU, try to use the OAuth2 authentication with the access key / secret under the project Id
+  if (!params.forceServicePlanIdUsageForSmsApi
+    && (region === Region.UNITED_STATES || region === Region.EUROPE)) {
```
JPPortier, line 45: typescript question for my own growth:
When calling `buildFlexibleOAuth2OrApiTokenApiClientOptions` with `region` parameter value as string `us` (and **not** `Region.UNITED_STATES`)
Is the test result will be as expected here ?

asein-sinch: It would be the same result as by assigning string values to enum members, Typescript infers these values as string literals, so in any case, we are comparing strings with strings

JPPortier (approved):

PR #54, opened 2024-04-05
DEVEXP-381: Flexible enums for regions

asein-sinch:


packages/conversation/tests/rest/v1/conversation-domain-api.test.ts
```diff
@@ -31,18 +31,15 @@ describe('Conversation API', () => {
...
-  //   const consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
-  //   conversationApi.getSinchClient();
-  //   // Add a small delay to allow jest to capture the warning
-  //   setTimeout(() => {
-  //     expect(consoleWarnSpy).toHaveBeenCalledWith('The region \'ca\' is not supported for the Conversation API');
-  //     consoleWarnSpy.mockRestore();
-  //   }, 20);
-  // });
+  it('should log a warning when using an unsupported region', async () => {
+    params.conversationRegion = 'bzh';
```
JPPortier, line 35: This is a very good region 😉

packages/conversation/src/rest/v1/conversation-domain-api.ts
```diff
@@ -86,13 +87,14 @@ export class ConversationDomainApi implements Api {
     return this.client;
   }
 
-  private buildHostname(region: ConversationRegion) {
+  private buildHostname(region: ConversationRegionFlexible) {
+    const formattedRegion = region === '' ? region : `${region}.`;
```
JPPortier, line 91: I think it come from myself, but thank you to small function usage here: my first reading of this line was cause of reflexion to understand logic and thinking there was a issue:
- testing against an empty value to use the testing value (the empty one)
- if not empty: use it too....

🤔 
Then I realized the final dot for second case; then use to format a string
Just a personal thought I think I would have written logic as 
```const formattedRegion = region !== '' ? `${region}.` '';```
Which could highlight in a better way the logic and use case.
Again just a personal thought and not a blocker

File: packages/sdk-client/src/domain/domain-interface.ts [hunk 3]
```diff
@@ -81,53 +83,34 @@ export enum SmsRegion {
...
-
-  for (const region of Object.values(SmsRegion)) {
-    if (region === value.toLowerCase()) {
-      return region as SmsRegion;
-    }
-  }
-  console.error(`No region exist for the value '${value}'`);
-  return undefined;
-}
+export type SmsRegionFlexible = SmsRegion | string;
```
JPPortier, line 86: By inverting the point of view; could the type being more user friendly ?
I mean:
```javascript
export enum SupportedSmsRegion {
  UNITED_STATES = 'us',
  EUROPE = 'eu',
  BRAZIL = 'br',
  CANADA = 'ca',
  AUSTRALIA = 'au'
}
export type SmsRegion = SupportedSmsRegion | string;
```
So when function prototype or variable usage will reference the type, it will be named `SmsRegion`.
Seems to be more "user friendly" than a `SmsRegionFlexible` reference and usage.
What do you think?

packages/conversation/src/rest/v1/conversation-domain-api.ts
asein-sinch: You are right, it would be more readable.

packages/sdk-client/src/domain/domain-interface.ts
asein-sinch: This would be more understandable indeed. The issue being that with this implementation, I can't do SmsRegion.EUROPE as SmsRegion refers to a Type, but would be used as a value
A solution would be to trick the user and define on top of the enum a new object SmsRegion such as 
```
export const SmsRegion = {
  ...SupportedSmsRegion,
};
```
WDYT?

JPPortier: As usual, we should have end user and SDK usage in mind first.
so if a `SmsRegion` type is identified as  the best way to reach this objective, if by implementing it as you described in your comment above: it is **very** fine
Then, using same pattern for Conversation, if SDK usage enable to store a variable/use a parameter type named `ConversationRegion` and reference `ConversationRegion.UNITED_STATES` or use `us` directly : it should provide a better DX (IMO)

asein-sinch: Ok, we can do that. Keep in mind the trick I mentioned above: ConversationRegion.UNITED_STATES will not reference the enum, but an object of which we read the value associated to the key `UNITED_STATES` 😉

asein-sinch: Fixed

JPPortier (approved):
👍

PR #55, opened 2024-04-09
DEVEXP-382_Change package exports for Fax API

asein-sinch:
This PR showcases a new way of exporting an API packages (The rationale behind the need to do that is explained in the Jira ticket)
Summary of the changes:
 - extract the requestData interfaces from the APIs files to put them in a dedicated folder /models/requests
 - them for the webhook event parsed
 - Breaking change:
   - export the models as `Fax`
   - update everywhere in the code where a model was imported
Before:
```typescript
import { AddEmailToNumbersRequestData } from '@sinch/sdk-core';
const requestData: AddEmailToNumbersRequestData = { ... }
``` 
Now:
```typescript
import { Fax } from '@sinch/sdk-core';
const requestData: Fax.AddEmailToNumbersRequestData = { ... }
``` 

If you agree with this approach, the same needs to be done for the other APIs. This PR is just a POC of a "small" API.


JPPortier:
I did not dug to full implementation but according to PR's description, what you're trying to fix could be compared to package tree for java to avoid name collisions.
So general question before details: have you look at migrating to the use of `namespace`?
That seems to be exactly what it's for: https://www.typescriptlang.org/docs/handbook/namespaces.html

asein-sinch:
> So general question before details: have you look at migrating to the use of namespace?

Thanks for your remark and for considering different approaches for organizing the TypeScript code. I appreciate your willingness to explore the use of namespaces and I'm open to discussing this further. 
I would like to share some considerations that have let me to prefer not using them: it's considered better practice to use the ES2015 module syntax (using `import` and `export` statements) as it aligns with industry best practices and offers several advantages in terms of compatibility, encapsulation, and optimization. It's also future-proofing the codebase and staying aligned with modern JavaScript development practices: it ensures the code remains compatible with evolving tooling and easier to onboard new team members.

JPPortier:
> > So general question before details: have you look at migrating to the use of namespace?
> 
> Thanks for your remark and for considering different approaches for organizing the TypeScript code. I appreciate your ....
👍

JPPortier (approved):

PR #56, opened 2024-04-09
DEVEXP-375: Enhance DX for ttl field in SendMessageRequest

asein-sinch:
This is the Node.js proposal about handling the `ttl` field for the SendMessageRequest in the Conversation API:
 - the field says what it is
 - the user just needs to give a value, the SDK takes care of the formatting

### DIFF
--- a/packages/conversation/src/models/v1/send-message-request/send-message-request.ts
+++ b/packages/conversation/src/models/v1/send-message-request/send-message-request.ts
@@ -36,7 +36,7 @@ export interface SendMessageRequestBase<T extends Recipient> {
   /** @see Recipient */
   recipient: T;
   /** The timeout allotted for sending the message, expressed in seconds. Passed to channels which support it and emulated by the Conversation API for channels without ttl support but with message retract/unsend functionality. Channel failover will not be performed for messages with an expired TTL. The format is an integer with the suffix `s` (for seconds). Valid integer range is 3 to 315,576,000,000 (inclusive). Example values include `10s` (10 seconds) and `86400s` (24 hours). */
-  ttl?: string;
+  ttl_seconds?: number;
   /** Overrides the app\'s [Processing Mode](../../../../../conversation/processing-modes/). Default value is `DEFAULT`. */
   processing_strategy?: ProcessingStrategy;
   /** An arbitrary identifier that will be propagated to callbacks related to this message, including MO replies. Only applicable to messages sent with the `CONVERSATION` processing mode. Up to 128 characters long. */
--- a/packages/conversation/src/models/v1/send-message-request/send-message-request.ts
+++ b/packages/conversation/src/models/v1/send-message-request/send-message-request.ts
@@ -35,9 +35,13 @@ export interface SendMessageRequestBase<T extends Recipient> {
   queue?: MessageQueue;
   /** @see Recipient */
   recipient: T;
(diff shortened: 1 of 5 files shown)


packages/conversation/src/models/v1/send-message-request/send-message-request.ts
```diff
@@ -36,7 +36,7 @@ export interface SendMessageRequestBase<T extends Recipient> {
   /** @see Recipient */
   recipient: T;
   /** The timeout allotted for sending the message, expressed in seconds. Passed to channels which support it and emulated by the Conversation API for channels without ttl support but with message retract/unsend functionality. Channel failover will not be performed for messages with an expired TTL. The format is an integer with the suffix `s` (for seconds). Valid integer range is 3 to 315,576,000,000 (inclusive). Example values include `10s` (10 seconds) and `86400s` (24 hours). */
-  ttl?: string;
+  ttl_seconds?: number;
```
JPPortier, line 39: My 2 cents
As a developer/SDK's user: Hiding at high level the underlying protocol is going to the right direction, so, setting a number for a period in place a string with a dedicated format : 👍 

For this dedicated context (`message.ttl field`) : may be we can **not** renaming the field itself.
What about 
```javascript
ttl_seconds?: number|string;
```
Then onto serialization (message-api.ts), something like:
```javascript
    // Special fields handling: ttl: number =>  string
    if (typeof data.sendMessageRequestBody.ttl == "number") {
     sendMessageRequestBody.ttl = `${data.sendMessageRequestBody.ttl}s`;
    }
```
So, SDK is accepting direct protocol format if required and also a number.

This do not help to define a general pattern to be common to all domains/type.
Having said this, renaming the field could be a good option too: if in the future protocol is accepting a period larger enough to not been contained in a `number` representation (theoretical hypothesis in this context `Number.MAX_VALUE=1.7976931348623157E+308` is a very long period to waiting for a message 😃) 

To summarize:
- not renaming the field: consistency with documentation and cross SDK
- renaming the field: open to the future but SHOULD/MUST be applied to SDK documentation then applied across all SDKs too

packages/conversation/src/rest/v1/messages/messages-api.ts
```diff
@@ -361,6 +361,12 @@ export class MessagesApi extends ConversationDomainApi {
       'Accept': 'application/json',
     };
 
+    // Special fields handling: ttl_seconds: number => ttl: string
+    if (data.sendMessageRequestBody.ttl_seconds !== undefined) {
+      (data as any).sendMessageRequestBody.ttl = data.sendMessageRequestBody.ttl_seconds + 's';
+      delete data.sendMessageRequestBody.ttl_seconds;
```
JPPortier, line 367: ⚠️ Very not fan of this: objects immutability
What  about a user trying to reuse same instance across different calls? 
After 1st serialization, the field is deleted and then following "send request" won't act as expected by user

packages/conversation/src/models/v1/send-message-request/send-message-request.ts
asein-sinch: This makes sense, I'll revert the name to `ttl` and adapt the transformation accordingly

packages/conversation/src/rest/v1/messages/messages-api.ts
asein-sinch: I'll create a copy of the object to make sure I don't mess with it.

JPPortier (approved):

asein-sinch:
Thanks @JPPortier for the review and the design solution. As the solution extends the basic specification, there is no blocker to merge this PR now !

PR #63, opened 2024-04-11
DEVEXP-382: Conversation API - Change package exports

asein-sinch:


JPPortier (approved):
Good to have "namespace" like usage (even for helpers)

asein-sinch:
Yes, and thanks (again!) for your comments, I didn't find it useful when I wrote the first version of the code, but I'm convinced it's a great addition for the developer experience to access the helpers under a dedicated namespace. It's far easier to find.

PR #71, opened 2024-04-25
Build on version branches too

asein-sinch:
Feature development is done on vX.Y+1-next
Bugfix is done on vX.Y

### DIFF
--- a/.github/workflows/run-ci.yaml
+++ b/.github/workflows/run-ci.yaml
@@ -1,11 +1,16 @@
-name: Test Sinch Node.js SDK
+name: Build and test Sinch Node.js SDK
 
 on:
   push:
     branches:
-      - 'main'
+      - main
+      - v([0-9]+)\.[0-9]+
+      - v([0-9]+)\.[0-9]+-next
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-ci.yaml
```diff
@@ -1,11 +1,16 @@
-name: Test Sinch Node.js SDK
+name: Build and test Sinch Node.js SDK
 
 on:
   push:
     branches:
-      - 'main'
+      - main
+      - v([0-9]+)\.[0-9]+
+      - v([0-9]+)\.[0-9]+-next
```
Dovchik, line 8: Can we add comments with examples what this regex resolves too?
```suggestion
      # matches e.g. 1.1 or 1.43
      - v([0-9]+)\.[0-9]+
      # matches e.g. 1.1-next
      - v([0-9]+)\.[0-9]+-next
```

asein-sinch: Thanks for your suggestion but this is not necessary, I've changed the configuration to build on any push

Dovchik (approved):

PR #73, opened 2024-04-26
Feature/Elastic SIP Trunking

asein-sinch:
Feature branch for all the work related to the Elastic SIP Trunking SDK development


packages/sdk-client/src/api/api-client.ts
```diff
@@ -5,8 +5,13 @@ import FormData = require('form-data');
 
 export enum PaginationEnum {
```
JPPortier, line 6: May be having such list + naming + comment + dedicated field names variation processing is highlighting a design to be improved: is it possible to have something like
- a common interface definition/abstraction (avoiding dedicated payload names)
- each domain is implementing it own real data to data abstraction service
 
And, because at domain level, it won't "hurt" to have domain logic moved  from `generic client` to `dedicated domain`

packages/sdk-core/README.md
```diff
@@ -166,17 +174,18 @@ const { SinchClient } = require('@sinch/sdk-core');
...
-| Voice          | voiceHostname                      | voiceRegion        |
-|                | voiceApplicationManagementHostname |                    |
-| Numbers        | numbersHostname                    | N/A                |
-| Verification   | verificationHostname               | N/A                |
+| API Name              | Parameter                          | Region             |
+|-----------------------|------------------------------------|--------------------|
+| Authentication        | authHostname                       | N/A                |
+| SMS                   | smsHostname                        | smsRegion          |
+| Conversation          | conversationHostname               | conversationRegion |
+|                       | conversationTemplatesHostname      |                    |
```
JPPortier, line 182: Adding `N/A` for both empty cells (conversationTemplatesHostname, voiceApplicationManagementHostname) could have sense)

packages/elastic-sip-trunking/README.md
```diff
@@ -0,0 +1,121 @@
...
+## Usage
+
+### Credentials
+
+The `Elastic SIP Trunking` API uses the Sinch unified authentication with OAuth2. You will need to provide the following credentials:
+- projectId: can be found in the [Account Dashboard](https://dashboard.sinch.com/settings/access-keys)
+- keyId:: can be found in your Access key list in the [Account Dashboard](https://dashboard.sinch.com/settings/access-keys)
+- keySecret: can be found **ONLY** when generating a new access key: keep it safe!
+
+### As part of the Sinch SDK
```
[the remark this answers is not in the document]
asein-sinch: Indeed, now that I promote the usage of the service instead of the Sinch Client, it really makes sense to factorize the documentation.
Thanks for the suggestion, I'll do it for all the APIs in a separate PR

packages/elastic-sip-trunking/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@sinch/elastic-sip-trunking",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
asein-sinch: It will be changed when preparing the release, but I can do it right away too.

packages/sdk-core/README.md
asein-sinch: Yes for `voiceApplicationManagementHostname`
No for `conversationTemplatesHostname`: I'll update it to `conversationRegion` as it is regionalized too and the same region must be used with the conversation API

PR #74, opened 2024-04-26
DEVEXP-416: New pagination type for EST lists

asein-sinch:


packages/sdk-client/src/api/api-client.ts
```diff
@@ -5,8 +5,13 @@ import FormData = require('form-data');
...
 export enum PaginationEnum {
   NONE,
+  /** Used by the Numbers API */
   TOKEN,
+  /** used by the SMS API */
   PAGE,
+  /** used by the Elastic SIP Trunking API */
+  PAGE2,
+  /** used by the Fax API */
   PAGE3
```
Dovchik, line 15: What can we tell from the names of the enum values? Maybe we can add descriptive name to not get confused later? e.g.
```suggestion
  INDEX_SMS_PAGE,
  /** used by the Elastic SIP Trunking API */
  ELASTIC_SIP_INDEX_PAGE,
  /** used by the Fax API */
  FAX_INDEX_PAGE
```

Dovchik (commented):
Is there unit test for the changes? I think this is a good candidate for coverage

packages/sdk-client/src/api/api-client.ts
asein-sinch: I hope that not all the APIs will use a different pagination and that the same pagination type can be used by several APIs as the OAuth2 authentication is used by several APIs

asein-sinch:
> Is there unit test for the changes? I think this is a good candidate for coverage

I haven't added a test for this new feature indeed. Will do.

asein-sinch:
@Dovchik : Tests added

packages/sdk-client/src/api/api-client.ts
Dovchik: Still it would nice to have a meaningful names to the enum then to clarify intention but that's up to you

Dovchik (approved):

PR #77, opened 2024-04-29
DEVEXP-411: Add support for Access control List

asein-sinch:


packages/elastic-sip-trunking/src/rest/v1/access-control-list/access-control-list-api.ts
```diff
@@ -0,0 +1,335 @@
...
+      'Accept': 'application/json',
+    };
+
+    const body: RequestBody = '';
+    const basePathUrl = `${this.client.apiClientOptions.hostname}/v1/projects/${this.client.apiClientOptions.projectId}/accessControlLists`;
+
+    const requestOptionsPromise = this.client.prepareOptions(basePathUrl, 'GET', getParams, headers, body || undefined);
+
+    const operationProperties: PaginatedApiProperties = {
+      pagination: PaginationEnum.PAGE2,
```
JPPortier, line 206: Not related to this PR (see comment from another PR related to pagination handling from client/core)
`PaginationEnum.PAGE2`, `PAGE3`, ... are not very intuitive

asein-sinch: Totally agree: at the beginning, there was only TOKEN and PAGE, but then I noticed that the parameters for the PAGE pagination are different from one API to the other and even the start page is not the same. So as Fax is v3, I thought the APIs were following a kind of evolution and I created a new pagination type called PAGE3.
And then I saw this one and I was out of ideas to find a name that would characterise the pagination behaviour. So I went for the opportunistic PAGE2 that was conveniently filling the gap 😅

PR #80, opened 2024-04-30
DEVEXP-409: Add support for Phone Numbers

asein-sinch:


Dovchik (approved):
For a note, the tests doesn't check for proper URL matching, passed headers and proper serialization/serialization, I think that service(and maybe all other) is a good candidate for proper e2e with doppelganger or other HTTP mock framework

asein-sinch:
> For a note, the tests doesn't check for proper URL matching, passed headers and proper serialization/serialization, I think that service(and maybe all other) is a good candidate for proper e2e with doppelganger or other HTTP mock framework

This is indeed the purpose of these unit tests: testing the exposed methods for a given domain and their input / output. Changes in the current exposed SDK API should fail the tests.

PR #83, opened 2024-05-06
Fix discrepancies from documentation

asein-sinch:
- The `updateTime` property will always be returned, either with the value `null`, either with a date value. Note that the server returns dates in UTC but without the timezone definition.
- The list of SIP trunks is returned in a property called `trunks` and not `sipTrunks` as we can read in the [documentation](https://developers.sinch.com/docs/est/api-reference/sip-trunking/#operation/getSipTrunks!c=200&path=sipTrunks&t=response). The information has been reported to the EST team

### DIFF
--- a/packages/elastic-sip-trunking/src/models/v1/sip-endpoint/sip-endpoint.ts
+++ b/packages/elastic-sip-trunking/src/models/v1/sip-endpoint/sip-endpoint.ts
@@ -21,5 +21,5 @@ export interface SipEndpoint {
   /** The date and time that the SIP endpoint was created. */
   createTime?: Date;
   /** The date and time that the SIP endpoint was last modified. */
-  updateTime?: Date;
+  updateTime?: Date | null;
 }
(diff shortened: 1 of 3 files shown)


packages/elastic-sip-trunking/src/models/v1/sip-endpoint/sip-endpoint.ts
```diff
@@ -21,5 +21,5 @@ export interface SipEndpoint {
   /** The date and time that the SIP endpoint was created. */
   createTime?: Date;
   /** The date and time that the SIP endpoint was last modified. */
-  updateTime?: Date;
+  updateTime?: Date | null;
```
JPPortier, line 24: According to PR's comment:

> The updateTime property will always be returned

Does it have sense to let `updateTime` as optional ?

asein-sinch: 🤷‍♂️
It's the same question for the property `createTime` defined above: one can assume that a SIP Endpoint object can be returned only after it has been created first but the documentation specifies it as optional still.
And there are also properties with default values such as `port`, `transport` and `enabled`. 
`id` and `sipTrunkId` are also noted as optional but in reality, they will also be always defined, the `id` for obvious reasons and the `sipTrunkId` because it's a mandatory parameter in order to create a SIP Endpoint

JPPortier (approved):

PR #94, opened 2024-05-21
DEVEXP-435: Support 'Accept-Language' header for SMS Verification

asein-sinch:


packages/verification/src/rest/v1/verifications/verifications-api.ts
```diff
@@ -254,6 +254,11 @@ export class VerificationsApi extends VerificationDomainApi {
       'Content-Type': 'application/json; charset=UTF-8',
       'Accept': 'application/json',
     };
+    if (data.startVerificationWithSmsRequestBody.smsOptions?.locale !== undefined) {
+      headers['Accept-Language'] = data.startVerificationWithSmsRequestBody.smsOptions?.locale;
```
JPPortier, line 258: NIT but question mark do not have added value here because checked just above

packages/verification/src/rest/v1/verifications/verifications-api.ts
```diff
@@ -254,6 +254,11 @@ export class VerificationsApi extends VerificationDomainApi {
...
+      // Remove the `locale` property from the body as it is used as a header parameter for the API call
+      delete data.startVerificationWithSmsRequestBody.smsOptions?.locale;
```
JPPortier, line 260: If we are deleting body filled by SDK user we could break it implementation (i.e. same object reused within a loop)

asein-sinch: Arf, you're right, I should have used a copy

packages/verification/src/rest/v1/verifications/verifications-api.ts
asein-sinch: Fixed

JPPortier (approved):

PR #100, opened 2024-07-01
DEVEXP-474: E2E Numbers/Available regions

asein-sinch:


JPPortier (approved):
Thank you, will be helpful to implement Java related part 👌

PR #103, opened 2024-07-03
DEVEXP-365: E2E Numbers/Callback configuration

asein-sinch:


packages/numbers/tests/rest/v1/callbacks/calback-configuration.steps.ts
```diff
@@ -0,0 +1,50 @@
...
+    error = e;
+  }
+});
+
+Then('the response contains the updated project\'s callback configuration', () => {
+  assert.equal(callbackConfiguration.projectId, '12c0ffee-dada-beef-cafe-baadc0de5678');
+  assert.equal(callbackConfiguration.hmacSecret, 'strongPa$$PhraseWith36CharactersMax');
+});
+
+Then('the response contains an error', () => {
```
JPPortier, line 45: Seems cucumber is able to select proper  `Then` to use for validation for differents `When` ?
Two when:
- https://github.com/sinch/sinch-sdk-mockserver/pull/7/files#diff-f6040bcb964ba27fd9629c90a7a2bdece0f9a789498cfab8a54c5aa13b8522fcR11
- https://github.com/sinch/sinch-sdk-mockserver/pull/7/files#diff-f6040bcb964ba27fd9629c90a7a2bdece0f9a789498cfab8a54c5aa13b8522fcR16

And here we are having two different `Then` validation for a single `When` definition

JPPortier (approved):

PR #105, opened 2024-07-05
DEVEXP-478: E2E Conversation/Messages

asein-sinch:
On top of adding steps for the Conversation / Messages, this PR fixes a bug in the pagination for the Conversation API: it was using the mode `TOKEN` designed for the Numbers API, but the pagination parameter are different in the Conversation API (snake_case) so I had to create the mode `TOKEN_2` to handle it properly.
Note: before commenting about `TOKEN_2` being not explicit enough, I want to highlight that it is not exposed to the end user and since the pagination is finally different for all the APIs integrated in the SDK, I'll have to re-think about the initial design in another ticket to create, refine and prioritise.

### DIFF
--- a/packages/conversation/tests/rest/v1/messages/messages.steps.ts
+++ b/packages/conversation/tests/rest/v1/messages/messages.steps.ts
@@ -1,0 +1,126 @@
+import { Conversation, ConversationService, MessagesApi, SupportedConversationRegion } from '../../../../src';
+import { Given, Then, When } from '@cucumber/cucumber';
+import * as assert from 'assert';
+import { PageResult } from '@sinch/sdk-client';
+
+let messagesApi: MessagesApi;
+let messageResponse: Conversation.SendMessageResponse;
+let listResponse: PageResult<Conversation.ConversationMessage>;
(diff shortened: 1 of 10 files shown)


packages/conversation/tests/rest/v1/messages/messages.steps.ts
```diff
@@ -0,0 +1,126 @@
...
+    messagesList.push(service);
+  }
+});
+
+When('I iterate manually over the messages pages', async () => {
+  listResponse = await messagesApi.list({
+    page_size: 2,
+  });
+  let reachedEndOfPages = false;
+  while (!reachedEndOfPages) {
```
JPPortier, line 67: What do you think about a test validating the number of pages read  in addition to number of messages received ?

asein-sinch: Ok, I'll add it

asein-sinch: Done

JPPortier (approved):

PR #124, opened 2024-08-19
DEVEXP-516: E2E SMS/Delivery Reports

asein-sinch:
Breaking change notice: the query parameter `recipient_msisdn ` has been renamed `phone_number` in the request data interface  to simplify the parameter comprehension.

Model update: 
 - the parameter `recipients` is marked as required in the [specification](https://gitlab.com/sinch/sinch-projects/enterprise-and-messaging/documentation/developer-experience/oas-documentation/-/blob/master/services/sms-rest-api/api-docs.yaml?ref_type=heads#L2388) but it is the case only with the query parameter `type=full` => it's now optional in the `MessageDeliveryStatus` interface.
 - the parameter `code` can now also be a number or a list of numbers instead of a string only

### DIFF
--- a/packages/sms/tests/rest/v1/delivery-reports/delivery-reports.steps.ts
+++ b/packages/sms/tests/rest/v1/delivery-reports/delivery-reports.steps.ts
@@ -28,7 +28,7 @@ When('I send a request to retrieve a summary SMS delivery report', async () => {
       'Delivered',
       'Failed',
     ],
-    code: '15,0',
+    code: [15, 0],
   };
   deliveryReport = await deliveryReportsApi.get(requestData);
 });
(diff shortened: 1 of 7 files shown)


packages/sms/tests/e2e/features/delivery-reports.feature
JPPortier, line 1: Seems this file should not been present

asein-sinch: Indeed, sorry for that. It's not present anymore now.

JPPortier (approved):

packages/sms/tests/rest/v1/delivery-reports/delivery-reports.steps.ts
```diff
@@ -24,6 +24,11 @@ Given('the SMS service "Delivery Reports" is available', () => {
 When('I send a request to retrieve a summary SMS delivery report', async () => {
   const requestData: Sms.GetDeliveryReportByBatchIdRequestData = {
     batch_id: '01W4FFL35P4NC4K35SMSBATCH1',
+    status: [
+      'Delivered',
+      'Failed',
+    ],
+    code: '15,0',
```
asein-sinch, line 31: This parameter can be improved: the spec defines a list but for a better DX, we may change it to a list or integers

PR #126, opened 2024-08-21
DEVEXP-518: E2E SMS/Groups

asein-sinch:
Notice: breaking change in the interface names:
 - `CreateGroupResponse` / `ReplaceGroupResponse` / `UpdateGroupResponse` / `GroupResponse` have all disappeared to become a single `Group` interface to reduce the confusion when using a response object from the SMS groups endpoints.

### DIFF
--- a/packages/sms/tests/rest/v1/groups/groups.steps.ts
+++ b/packages/sms/tests/rest/v1/groups/groups.steps.ts
@@ -104,7 +104,7 @@ When('I send a request to update an SMS group', async () => {
   group = await groupsApi.update({
     group_id: '01W4FFL35P4NC4K35SMSGROUP1',
     updateGroupRequestBody: {
-      name: null,
+      name: 'Updated group name',
       add: [
         '+12017771111',
         '+12017772222',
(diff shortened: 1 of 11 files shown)


packages/sms/tests/rest/v1/groups/groups.steps.ts
```diff
@@ -0,0 +1,177 @@
...
+      ],
+      add_from_group: '01W4FFL35P4NC4K35SMSGROUP2',
+      remove_from_group: '01W4FFL35P4NC4K35SMSGROUP3',
+    },
+  });
+});
+
+Then('the response contains the updated SMS group details', () => {
+  assert.equal(group.id, '01W4FFL35P4NC4K35SMSGROUP1');
+  assert.equal(group.name, undefined);
```
JPPortier, line 124: This is exactly the purpose of the comment onto the feature PR's description
We need to ensure test is pointing this dedicated behavior: name field is missing not an empty string

asein-sinch: Fixed

JPPortier (approved):

PR #128, opened 2024-08-23
DEVEXP-520: E2E Verification/Start

asein-sinch:
On top of the implementation of the e2e steps, this PR starts the refactoring for the new naming convention regarding the "phone call" and "data" verifications.
I tried to make the code backward compatible by deprecating the references to the "callout" and "seamless" interfaces.
The refactoring for the reportVerification endpoint will take place with the [DEVEXP-521](https://tickets.sinch.com/browse/DEVEXP-521) user story


examples/integrated-flows-examples/src/verification/app.ts
```diff
@@ -18,9 +18,9 @@ dotenv.config();
 
   enum VerificationMethods {
     SMS ='sms',
-    CALLOUT = 'callout',
+    PHONE_CALL = 'phone call',
```
JPPortier, line 21: Underlying identifiers (protocol layer) are still 'callout' and 'seamless' (see https://gitlab.com/sinch/sinch-projects/enterprise-and-messaging/documentation/developer-experience/oas-documentation/-/blob/master/services/verification-oas/verification.yaml?ref_type=heads#L776)
Only the the _presenter_ terms was changed (no underlying protocol breaking change)

So we could imagine having:
```
/**
 * @deprecated The identifier should not be used and prefer PHONE_CALL usage
 */
CALLOUT = 'callout',
PHONE_CALL = 'callout',
```
(same for `SEAMLESS`)
So we avoid a hard break change and are able to provide a smooth transition period to users until the identifiers will be really removed (and will release SDK with a major version change i.e. : 2.0)

examples/integrated-flows-examples/src/verification/app.ts
```diff
@@ -41,21 +41,21 @@ dotenv.config();
       case VerificationMethods.SMS:
         startSmsVerificationFlow(answers.phoneNumber);
         break;
-      case VerificationMethods.CALLOUT:
-        startCalloutVerificationFlow(answers.phoneNumber);
+      case VerificationMethods.PHONE_CALL:
```
JPPortier, line 44: If we are moving to support smooth transition with both `CALLOUT` and `PHONE_CALL`; well need to duplicate the `case ` here:
```
case VerificationMethods.CALLOUT
case VerificationMethods.PHONE_CALL:
  ...
```

examples/integrated-flows-examples/src/verification/app.ts
asein-sinch: In this small program, I don't consider it as a breaking change but rather a bug fix to present the same terminology as in the documentation to the end user when the program will list the available verification methods. WDYT?

JPPortier: My bad,
I was not focused enough and did not take attention it is a sample app
I was thinking it was the "official" supported values for SDK
Sorry you should ignore this thread...

packages/verification/README.md
```diff
@@ -87,19 +90,22 @@ const credentials: ApplicationCredentials = {
...
 // Build the request data
-const requestData: Verification.StartVerificationRequestData = {
-  initiateVerificationRequestBody: {
+const requestData: Verification.StartSmsVerificationRequestData = {
+  startVerificationWithSmsRequestBody: {
     identity: {
       type: 'number',
       endpoint: '+17813334444',
     },
-    method: 'sms',
```
JPPortier, line 96: ⭐

packages/verification/README.md
```diff
@@ -87,19 +90,22 @@ const credentials: ApplicationCredentials = {
...
+      locale: 'sv-SE',
+    },
   },
 };
 
 // Use the standalone declaration of the 'verification' service
-const verificationInitResponse: Verification.InitiateVerificationResponse 
-    = await verificationService.verifications.start(requestData);
+const startVerificationResponse: Verification.StartSmsVerificationResponse 
+    = await verificationService.startVerifications.startSms(requestData);
```
JPPortier, line 108: Not fan of renaming `verificationService.verifications`  to `verificationService.startVerifications`
It could have sense ( with a major breaking change), but it start to drift with other SDK's provided implementation and ID <> SDKs function mapping

asein-sinch: Indeed, I generated from the OAS tag, I'll update according to the confluence page

JPPortier (approved):

PR #129, opened 2024-08-26
DEVEXP-521: E2E Verification/Report

asein-sinch:


packages/verification/src/rest/v1/verifications/verifications-api.ts
```diff
@@ -211,10 +266,59 @@ export class VerificationsApi extends VerificationDomainApi {
...
+   */
+  public async reportPhoneCallByIdentity(
+    data: ReportPhoneCallVerificationByIdentityRequestData,
+  ): Promise<PhoneCallVerificationReportResponse> {
+    this.client = this.getSinchClient();
+    (data.reportPhoneCallVerificationByIdentityRequestBody as any).method = 'callout';
+    const getParams = this.client.extractQueryParams<ReportPhoneCallVerificationByIdentityRequestData>(
+      data, [] as never[]);
+    const headers: { [key: string]: string | undefined } = {
+      'Content-Type': 'application/json; charset=UTF-8',
```
JPPortier, line 282: outside of the PR scope, but:
1. Verification OAS file do not specify any charset, only 'application/json' and is compliant with OAS (https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-10)
2. 'application/json' by default mean UTF-8 (RFC chapter  "3. Encoding": https://www.ietf.org/rfc/rfc4627.txt)

JPPortier (approved):

packages/verification/src/rest/v1/verifications/verifications-api.ts
asein-sinch: Indeed, there has been a spec update from the time these files were generated.

PR #131, opened 2024-08-28
DEVEXP-523: E2E Verification/Callbacks

asein-sinch:


packages/verification/tests/rest/v1/callbacks/webhooks-events.steps.ts
```diff
@@ -0,0 +1,82 @@
...
+let event: Verification.VerificationCallbackEvent;
+let formattedHeaders: IncomingHttpHeaders;
+
+const processEvent = async (response: Response) => {
+  formattedHeaders = {};
+  response.headers.forEach((value, name) => {
+    formattedHeaders[name.toLowerCase()] = value;
+  });
+  rawEvent = await response.text();
+  rawEvent = rawEvent.replace(/\s+/g, '');
```
JPPortier, line 17: `rawEvent` is no more raw 🤔 
Is authentication is failingif using the true 'raw' coming from `response.text()`?

packages/verification/tests/rest/v1/callbacks/webhooks-events.steps.ts
```diff
@@ -0,0 +1,82 @@
...
+});
+
+Then('the Verification event describes a "Verification Request" event type', () => {
+  const verificationRequestEvent = event as Verification.VerificationRequestEvent;
+  assert.equal(verificationRequestEvent.id, '1ce0ffee-c0de-5eed-d00d-f00dfeed1337');
+  assert.equal(verificationRequestEvent.event, 'VerificationRequestEvent');
+  const smsVerificationMethod: Verification.MethodEnum = 'sms';
+  assert.equal(verificationRequestEvent.method, smsVerificationMethod);
+  const identity: Verification.Identity = {
+    verified: false,
```
JPPortier, line 48: I understand mockerserver is returning a response-like server; then it return the un-specified `verified` field
But from an e2e point of view, do we need to ensure validation of this dedicated un-specified field ?

asein-sinch: Right, removed from assertions

packages/verification/tests/rest/v1/callbacks/webhooks-events.steps.ts
asein-sinch: Yes, this is a compromise I did to keep a nice readability for the mockserver JSON files: the body comes with the formatting (spaces and carriage returns), which is not what the server will actually return. This function ensures the data is the same as the one we will get on a webhook before it's being processed.

JPPortier: No way to have mockerserver sources containing readable JSON payload but stripped from spaces before sending them back?
So we could have both:
- nice JSON
- server-like responses sent back

asein-sinch: After some tries, I noticed that despite the response is minified in the mockserver config, the mockserver will send pretty print it before sending it on the network.
As on the SDK side, we apply strictly the algorithm, I'll update the signatures on the mockserver side and this line won't be necessary anymore: in the end, the data will be prettified all over the places: WYSIWYG

asein-sinch: Removed with the commit [b4f61d7](https://github.com/sinch/sinch-sdk-node/pull/131/commits/b4f61d7e1091bde10c5c0975abfc770953db9494#diff-454750f6a252b89d1c7ef0eae82fa7d76faaaa197ade4a8ef417a3dbc029cac1L17)

JPPortier (approved):

PR #134, opened 2024-09-05
DEVEXP-525: E2E Voice/Calls

asein-sinch:


packages/voice/tests/rest/v1/calls/calls.steps.ts
```diff
@@ -0,0 +1,118 @@
...
+  const participant: Voice.Participant = {
+    type: 'Number',
+    endpoint: '+12017777777',
+  };
+  assert.deepEqual(callInformation.to, participant);
+  assert.equal(callInformation.domain, 'pstn');
+  assert.equal(callInformation.duration, 14);
+  assert.equal(callInformation.status, 'FINAL');
+  const result: Voice.ResultEnum = 'ANSWERED';
+  assert.equal(callInformation.result, result);
```
Dovchik, line 37: nit
shouldn't IDE suggest an enum value in assertion, so this explicit binding is not really adds readability?
```suggestion
  assert.equal(callInformation.result, 'ANSWERED');
```

packages/voice/tests/rest/v1/calls/calls.steps.ts
```diff
@@ -0,0 +1,118 @@
...
+          'en-US',
+        ),
+      ],
+      action: Voice.svamlActionHelper.buildHangup(),
+    },
+  });
+});
+
+Then('the update call response contains no data', () => {
+  assert.deepEqual(updateCallResponse, {});
```
Dovchik, line 70: I was expecting `null` or `undefined`, not an empty object. Nevertheless, some our api endpoints return `{}` empty json, is this a case here?

packages/voice/tests/rest/v1/calls/calls.steps.ts
asein-sinch: This is done on purpose, in order to break the transpilation in case the enum changes. The assert mechanism would not detect it otherwise.

packages/voice/tests/rest/v1/calls/calls.steps.ts
asein-sinch: This assertion is not about the API result but the SDK result. If the response is empty, the SDK will return an empty object.

Dovchik (approved):
questions answered 👌 
BTW, don't forget to hit `rerequest` review from the reviewer if all questions answered
this is much more strong indicator for me to rereview then just new comments

PR #144, opened 2024-10-03
DEVEXP-592: Add workflow for publication

asein-sinch:


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,44 @@
...
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Build packages
+        run: yarn run build
+
+      - name: Run tests
+        run: yarn run test
```
JPPortier, line 30: Can we imagine having e2e test before publishing ?

asein-sinch: We can in this temporary setup.
The end game would be to trigger this workflow after the CI when there is a merge on the branch `main` (hence after a build, unit tests and E2E).
I'll update this workflow for the moment with the E2E.

asein-sinch: Done

JPPortier (approved):

PR #146, opened 2024-10-03
DEVEXP-595: [Bugfix] Remove scheduledProvisioning elements from updat…

asein-sinch:
…e number request


JPPortier (approved):
Good catch.
OAS file should define this field as readOnly
Created related tickets to fix specs (DEVEXP-596)

PR #148, opened 2024-10-11
DEVEXP-607: Mailgun Authentication

asein-sinch:


packages/sdk-client/src/domain/domain-interface.ts
```diff
@@ -32,6 +33,13 @@ export interface UnifiedCredentials {
   conversationRegion?: ConversationRegion;
 }
 
+export interface MailgunCredentials {
+  /** Your API Key created from the [Mailgun Dashboard](https://app.mailgun.com/settings/api_security) */
+  mailgunApiKey: string;
+  /** The region for the Mailgun API. Default region is empty */
```
JPPortier, line 39: Could have sense to have details about "default value is empty" and "default region is ..".

asein-sinch: Indeed. Fixed

JPPortier (approved):

PR #150, opened 2024-11-13
Mailgun release - DO NOT MERGE

asein-sinch:
This is the target implementation of what should be generated from the OAS and the configuration.
There is a change in the API design, where the path parameters are now part of the request parameters and not in the request data object.
The models are also declined in a SDK version and an API version with associated transformer methods.


packages/mailgun/tests/rest/v1/emails/emails-api.test.ts
```diff
@@ -0,0 +1,109 @@
...
+} from '../../../models/v1/emails/response';
+
+describe('EmailsApi', () => {
+  let emailsApi: EmailsApi;
+  let fixture: EmailsApiFixture;
+  let credentials: SinchClientParameters;
+
+  beforeEach(() => {
+    fixture = new EmailsApiFixture();
+    credentials = {
```
JPPortier, line 21: Should not be used/useful

packages/mailgun/src/rest/v1/emails/emails-api.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Delete scheduled and undelivered mail
+   * Deletes all scheduled and undelivered mail from the domain queue.  This endpoint must be called on the storage API host and in the domain's region.  e.g. https://storage-us-east4.api.mailgun.net/v3/example.com/envelopes
+   *
+   * The storage hosts are `storage-us-east4.api.mailgun.net`, `storage-us-west1.api.mailgun.net`, and `storage-europe-west1.api.mailgun.net`.
+   * @param { string } domainName - The name of the domain you want to delete envelope from
+   */
+  public async purgeDomainQueues(domainName: string): Promise<GenericResponse> {
+    const requests = this.storageHostnames.map((hostname) =>
```
JPPortier, line 128: We should find a better way to not send multiple request and not sending request outside of EU if located EU.
When https://documentation.mailgun.com/docs/mailgun/api-reference/openapi-final/tag/Events/#tag/Events/operation/get-v3-domain_name-events will be supported we could reduce to a single URL call

examples/simple-examples/package.json
```diff
@@ -184,10 +189,12 @@
   },
   "dependencies": {
     "@sinch/sdk-core": "^1.2.0",
-    "dotenv": "^16.3.1"
+    "dotenv": "^16.3.1",
+    "nodemailer": "^6.9.16"
```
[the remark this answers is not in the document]
asein-sinch: Here, we are in the examples part of the repository, and the idea is to showcase how a SDK user would use the product in their own project: I followed the same advice as the one given by the Mailgun team here: https://github.com/mailgun/mailgun.js/issues/144#issuecomment-979128267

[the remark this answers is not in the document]
JPPortier: My bad.
I was not focused enough and did not notice this package.json is the one for examples

examples/simple-examples/src/mailgun/emails/getEmail.ts
```diff
@@ -0,0 +1,27 @@
...
+  console.log('* GetEmail *');
+  console.log('************');
+
+  const domainName = getMailgunDomainFromConfig();
+  const storageKey = getMailgunStorageKeyFromConfig();
+
+  const mailgunService = initMailgunService();
+  let response;
+  try {
+    response = await mailgunService.emails.getEmail(domainName, storageKey);
```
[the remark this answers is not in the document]
asein-sinch: Thinking about it, I would prefer `getStoredEmail` and keep `get` for methods where the `id` is the same as the one created with the `create` method.
The resource that we are fetching here is totally different to what has been sent in the `send` method.

packages/mailgun/src/models/v1/emails/request/override-properties/override-properties.ts
```diff
@@ -0,0 +1,34 @@
...
+  trackingPixelLocationTop?: string;
+  /** Used to specify an IP Address to send an email that is owned by your account */
+  sendingIp?: string;
+  /** If an IP Pool ID is provided, the email will be delivered with an IP that belongs in that pool */
+  sendingIpPool?: string;
+  /** If set to `true` or `yes` this requires the message only be sent over a TLS connection.  If a TLS connection can not be established, Mailgun will not deliver the message.  If set to `false` or `no`, Mailgun will still try and upgrade the connection, but if Mailgun cannot, the message will be delivered over a plaintext SMTP connection.  The default is `false` */
+  requireTls?: string;
+  /** If set to `true` or `yes`, the certificate and hostname of the resolved MX Host will not be verified when trying to establish a TLS connection. If set to `false` or `no`, Mailgun will verify the certificate and hostname. If either one can not be verified, a TLS connection will not be established. The default is `false` */
+  skipVerification?: string;
+  /** Enables sending in test mode. Pass `yes` if needed.  See **Sending in Test Mode** */
```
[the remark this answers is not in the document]
asein-sinch: Even if the specification mentions `yes` as the only accepted value, I will expose a boolean to the user for a better user experience (BTW: the server handles `true` and `false` properly)

packages/mailgun/src/models/v1/emails/request/send-email-request/send-email-request.ts
```diff
@@ -0,0 +1,166 @@
...
+  html?: string;
+  /** AMP part of the message.  Please follow Google guidelines to compose and send AMP emails */
+  amp_html?: string;
+  /** Body of the message (text version) */
+  text?: string;
+  /** Name of a template stored via template API to use to render the email body.  See **Templates** for more information */
+  template?: never;
+} & CommonEmailProperties;
+
+export type MessageContentWhereHtmlContentCanBeFromTemplateOnly = {
```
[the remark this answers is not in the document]
asein-sinch: These ones have been a headache and I couldn't come with a better name to express the following:
 - `MessageContentWhereHtmlContentCanBeInlineOnly`: the user can use a `text`, a `amp-html` and/or a `html` property, each one creating a different content type (`html` creating a text/html)
  - `MessageContentWhereHtmlContentCanBeFromTemplateOnly`: the user can use a `text`, a `amp-html` and/or a `template` property, each one creating a different content type (`template` creating a text/html)
  
So a user can create an object of type MessageContentWhereHtmlContentCanBeFromTemplateOnly with the property `text` only, making it weird to speak about a "TemplateEmail" in the type name.
```typescript
const email: SendTemplateEmailRequest = {
  text: 'Text content',
};
```

So with the name `MessageContentWhereHtmlContentCanBeInlineOnly`, it expresses that when this type is used, the property corresponding to the text/html content type can be written inline only in the `html` property.
And with the name `MessageContentWhereHtmlContentCanBeFromTemplateOnly`, it expresses that if the SDK user wants to render some HTML in the email body, the only possibility will be to use a predefined template.
I wish would could find more concise names to express the nature of these interfaces. Let me know if you find better alternatives considering all the points I've mentioned here.

packages/mailgun/src/models/v1/emails/request/send-email-request/send-email-request.ts
```diff
@@ -0,0 +1,166 @@
+import { TemplateProperties } from '../template-properties';
+import { OverrideProperties } from '../override-properties';
+import FormData = require('form-data');
+
+export type MessageContentWhereHtmlContentCanBeInlineOnly = {
```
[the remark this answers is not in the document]
asein-sinch: See the comment on the other type name ⬇️

packages/mailgun/src/models/v1/emails/request/send-email-request/send-email-request.ts
```diff
@@ -0,0 +1,166 @@
...
+  amp_html?: string;
+  /** Body of the message (text version) */
+  text?: string;
+  /** Name of a template stored via template API to use to render the email body.  See **Templates** for more information */
+  template?: never;
+} & CommonEmailProperties;
+
+export type MessageContentWhereHtmlContentCanBeFromTemplateOnly = {
+  /** Name of a template stored via template API to use to render the email body.  See **Templates** for more information */
+  template?: string;
```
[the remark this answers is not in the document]
asein-sinch: Right, fixed.

packages/mailgun/src/models/v1/emails/request/send-email-request-base/send-email-request-base.ts
```diff
@@ -0,0 +1,23 @@
+import { OverrideProperties } from '../override-properties';
+import { TemplateProperties } from '../template-properties';
+
+export interface SendEmailRequestBase {
```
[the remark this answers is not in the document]
asein-sinch: Indeed. Removed.

packages/mailgun/src/models/v1/emails/response/get-email-response/get-email-response.ts
```diff
@@ -0,0 +1,58 @@
+export const transformGetEmailResponseIntoClientResponse = (apiResponse: GetEmailResponseFromApi): GetEmailResponse => {
+  return {
+    contentTransferEncoding: apiResponse['Content-Transfer-Encoding'],
+    contentType: apiResponse['Content-Type'],
+    messageId: apiResponse['Message-Id'],
+    mimeVersion: apiResponse['Mime-Version'],
+    to: apiResponse['To'],
+    deliveryTime: apiResponse['X-Mailgun-Deliver-By'],
```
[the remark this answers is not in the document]
asein-sinch: In the specification no. In the server response, yes!

packages/mailgun/tests/rest/v1/emails/emails-api.test.ts
asein-sinch: Fixed

packages/mailgun/src/rest/v1/emails/emails-api.ts
asein-sinch: Thinking about it, I'm not sure this method will be helpful as we must be sure is stores all the events related to the scheduled emails long enough and the resulting operation can take some time, depending on the number of pages to fetch, the availability of the API, the network speed...
With the approach I have implemented, we don't care about any of that.

examples/simple-examples/src/mailgun/emails/getEmail.ts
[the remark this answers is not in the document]
JPPortier: 👍 `getStoredEmail` is fine

JPPortier (approved):
Approved and according to our discussion we will need to think again about naming before releasing/GA

packages/sdk-client/src/domain/domain-helper.ts
```diff
@@ -6,6 +6,11 @@ export const CONVERSATION_TEMPLATES_HOSTNAME = `https://${REGION_PATTERN}templat
 export const ELASTIC_SIP_TRUNKING_HOSTNAME = 'https://elastic-trunking.api.sinch.com';
 export const FAX_HOSTNAME = `https://${REGION_PATTERN}fax.api.sinch.com`;
 export const MAILGUN_HOSTNAME = `https://api.${REGION_PATTERN}mailgun.net`;
+export type MailgunStorageHostname =
```
JPPortier, line 9: Onto next iteration, an added value to have these enum values coming from generator: they could be shared across SDKs

packages/mailgun/src/models/v1/emails/request/override-properties/override-properties.ts
```diff
@@ -0,0 +1,34 @@
+export interface OverrideProperties {
+  /** Tag string.  See **Tagging** for more information */
+  tag?: string;
+  /** Toggles Send Time Optimization (STO) on a per-message basis.  String should be set to the number of hours in `[0-9]+h` format, with the minimum being `24h` and the maximum being `72h`.  This value defines the time window in which Mailgun will run the optimization algorithm based on prior engagement data of a given recipient.  See **Sending a Message with STO** for details.  *Please note that STO is only available on certain plans. See www.mailgun.com/pricing for more info* */
+  deliveryTimeOptimizePeriod?: string;
```
JPPortier, line 5: Added value manage a simple integer, like what was done with conversation for `ttl` field, and serialized it according to server expectation : https://github.com/sinch/sinch-sdk-node/blob/main/packages/conversation/src/rest/v1/messages/messages-api.ts#L297

packages/mailgun/src/models/v1/emails/request/send-email-request/send-email-request.ts
JPPortier, line 1: For DX, a set of fields could defined as simple bool
- https://github.com/sinch/sinch-sdk-java/pull/173/files#diff-84364d4dd8c32c0aca38015bef3f602d1be590187cacf0812f66e16e8555bb3c
- https://github.com/sinch/sinch-sdk-java/pull/173/files#diff-bfb89b4287e2d4bb6511b5d35a22289b7d7985527b821a71c2c7e4a2e46f2ebe

packages/mailgun/src/rest/v1/emails/emails-api.ts
```diff
@@ -0,0 +1,193 @@
...
+   * Deletes all scheduled and undelivered mail from the domain queue.  This endpoint must be called on the storage API host and in the domain's region.  e.g. https://storage-us-east4.api.mailgun.net/v3/example.com/envelopes
+   *
+   * The storage hosts are `storage-us-east4.api.mailgun.net`, `storage-us-west1.api.mailgun.net`, and `storage-europe-west1.api.mailgun.net`.
+   * @param { string } domainName - The name of the domain you want to delete envelope from
+   * @param { MailgunStorageHostname } storageHostname - The storage hostname to be purged
+   */
+  public async purgeSendingQueue(
+    domainName: string,
+    storageHostname: MailgunStorageHostname,
+  ): Promise<GenericResponse> {
```
JPPortier, line 136: What do you think about having a `void` return value ?
GenericResponse do not have added value over the HTTP 200 return code, or there is a need to define the expected message within response body to clearly define if operation was fine or not (and the `done` string as expected response is not part of specification)
In this context a simple expectation to throw an expectation if error with a void function seems to simplify the interface with SDK

packages/mailgun/src/models/v1/emails/request/send-email-request/send-email-request.ts
asein-sinch: It will be a custom type with `boolean | 'yes' | 'no'` to ensure a minimum compatibility with the existing mailgun.js SDK

packages/mailgun/src/rest/v1/emails/emails-api.ts
asein-sinch: Indeed, this makes sense for a DELETE operation

asein-sinch: done

JPPortier:
Are you sure for destination branch ?

PR #151, opened 2024-12-12
DEVEXP-642: Update types and refactor transformers

asein-sinch:


packages/mailgun/src/models/v1/emails/request/helper.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const appendFilteredPropertiesToFormData = (
+  obj: Record<string, any>,
+  prefix: string,
+  formData: FormData,
+) => {
+  for (const [key, value] of Object.entries(obj)) {
+    if (key.startsWith(prefix) && value != null) {
+      if (value instanceof Date) {
+        formData.append(key, value.toISOString());
```
JPPortier, line 11: Mailgun servers won't expect this format
- mailgun: https://documentation.mailgun.com/docs/mailgun/api-reference/openapi-final/tag/Messages/#tag/Messages/operation/POST-v3--domain-name--messages!path=o:deliverytime&t=request
- javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

packages/mailgun/src/models/v1/emails/request/send-mime-email-request/send-mime-email-request.ts
```diff
@@ -23,91 +24,22 @@ export interface SendMimeEmailRequest {
...
   const formData = new FormData();
-  if (sdkRequest['to'] !== undefined && sdkRequest['to'] !== null) {
+  if (sdkRequest['to'] != null) {
     formData.append('to', sdkRequest['to']);
   }
-  if ('message' in sdkRequest && sdkRequest['message'] !== undefined && sdkRequest['template'] !== null) {
-    formData.append('message', sdkRequest['message'], {
-      filename: 'MimeMessage',
-    });
+  if ('message' in sdkRequest && sdkRequest['message'] != null) {
```
JPPortier, line 30: All `emai` and some of `mime-emai`l fields are validated simple with simplified  `if (sdkRequest['XXX'] != null) {`
But not here (nor `template`) field.
Is it by purpose ?

packages/mailgun/src/models/v1/emails/request/template-properties/template-properties.ts
```diff
@@ -1,8 +1,23 @@
...
+
+export const appendTemplatePropertiesToFormData = (templateProperties: TemplateProperties, formData: FormData) => {
+  if (templateProperties['text'] != null) {
+    formData.append('t:text', String(templateProperties['text']));
+  }
+  if (templateProperties['version'] != null) {
+    formData.append('t:version', templateProperties['version']);
+  }
+  if (templateProperties['variables'] != null) {
+    formData.append('t:variables', templateProperties['variables']);
```
JPPortier, line 21: Not for further improvements and for all SDKs: we should imagine changing the type and provide an helper for SDKs users helping them serialize to JSON automaticcaly

packages/mailgun/src/models/v1/emails/request/send-email-request/index.ts
```diff
@@ -1 +1,2 @@
 export type { SendEmailRequest } from './send-email-request';
+export { transformSendEmailRequestIntoApiRequestBody } from './send-email-request';
```
JPPortier, line 2: All these function become public and outside of SDK internal usage scope.
This could cause namespace confusion/pollution.
Is there a way to reduce their scope/visibility to SDK internals only, and then reduce risk to have users with breaking change in case of internal changes in the future ?

packages/mailgun/src/models/v1/emails/request/helper.ts
asein-sinch: Here we are in the free-form variables and a choice has to be made in case the SDK user inputs a Date object. I chose ISO8601 but you're right, let's be consistent with he rest of the API

packages/mailgun/src/models/v1/emails/request/send-mime-email-request/send-mime-email-request.ts
asein-sinch: The check of the property existence was indeed a remain from the past. I've deleted it and updated the property check with the loose inequality operator `!=` (checks for value equality with type coercion)

packages/mailgun/src/models/v1/emails/request/send-email-request/index.ts
asein-sinch: There is no risk of namespace confusion as all the exports inside the `models` package are exported through a namespace. 
Now, regarding the risk to expose the method externally, due to this barrel file the method is indeed exposed on the Mailgun namespace and can be used this way:
```
import { Mailgun } from '@sinch/sdk-core';
Mailgun.transformSendEmailRequestIntoApiRequestBody({ foo: 'bar' });
```

There are several way to circumvent the issue, more or less impactful and more or less working:
 - option 1: remove this export from the barrel file. The method is still exposed and a warning can be added to explicitly document this is an internal file that should not be used and can be renamed at any time. In order to use it, a SDK user would have to dig into the whole repository arborescence, such as
```
import {
  transformSendEmailRequestIntoApiRequestBody
} from '@sinch/mailgun/src/models/v1/emails/request/send-email-request';
```
 - option 2: remove this method from the model file and add it inline in the emails-api.ts file, along with the API call. This solves completely the issue but will bring some additional complexity to test it. Indeed, the test will have to be written in the same file, such as 
 ```
 if (process.env.NODE_ENV === 'test') {
  describe('transformSendEmailRequestIntoApiRequestBody', () => {
    test('should transform email request correctly', () => {
      const result = transformSendEmailRequestIntoApiRequestBody({ foo: 'bar' });
      expect(result.body).toBe(JSON.stringify({ foo: 'bar' }));
    });
  });
}
 ```
 On top of polluting the src file, this will increase the size of the final bundle as this code will be present in the NPM package.
 
  - option 3: extract the transform code in other files with a specific pattern such as `send-email-request.transform.ts`, write a build script that will inject the code of the transformers method in the right place in the emails-api.ts file and exclude the *.transform.ts files from being part of the NPM package. This way, the method is still testable and becomes private to the SDK user. This approach is the most conmplex and most costly to put in place.

Let's discuss about the 3 options during a standup to decide which one to choose.

asein-sinch: Team decision => let's go for option 1

JPPortier (approved):

PR #152, opened 2024-12-12
Update codeowners list

asein-sinch:


JPPortier (approved):
approved but e2e failures not related to this PR to be investigated

asein-sinch:
The latest cucumber version still has a dependency on the vulnerable version of `cross-spawn`
The latest version of eslint is safe but the migration to a flat config is big enough to be done in a dedicated ticket: https://tickets.sinch.com/browse/DEVEXP-694

PR #153, opened 2024-12-12
Add missing content-type header when sending formdata

asein-sinch:
While working with Mailgun API and their formdata request, I noticed the content-type header was missing for Fax


JPPortier (approved):
approved but e2e failures to be investigated

asein-sinch:
Closing the PR: after investigation, the fetch lib computes the content-type header by itself as "multipart/form-data;boundary=...."

PR #167, opened 2025-05-19
DEVEXP-703: Support webhooks headers validation for SMS

asein-sinch:


packages/sms/src/rest/v1/callbacks/callbacks-webhook.ts
```diff
@@ -1,17 +1,31 @@
...
+  constructor(appSecret?: string) {
+    this.appSecret = appSecret;
+  }
+
   public validateAuthenticationHeader(
+    headers: IncomingHttpHeaders,
+    body: any,
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    _headers: IncomingHttpHeaders, _body: any, _path: string, _method: string,
+    _path?: string, _method?: string,
```
JPPortier, line 19: This is something I postponed myself: going to an unified function prototype across domains even if parameters are not used ❤️

packages/sms/tests/rest/v1/callbacks/webhooks-events.steps.ts
```diff
@@ -1,18 +1,24 @@
...
 import assert from 'assert';
+import { IncomingHttpHeaders } from 'http';
 
 let smsCallbackWebhook: SmsCallbackWebhooks;
 let rawEvent: any;
 let event: SmsCallback;
+let formattedHeaders: IncomingHttpHeaders;
 
 const processEvent = async (response: Response) => {
+  formattedHeaders = {};
```
JPPortier, line 12: This transformation should be part of SDK itself and not done only for end to end tests enabling to pass them here.

Seems this transformation is processed from SDK: https://github.com/sinch/sinch-sdk-node/blob/v1.3-next/packages/sdk-client/src/utils/authorization.helper.ts#L142

No ?

asein-sinch: Correct. Fixed now

JPPortier (approved):

PR #170, opened 2025-05-26
DEVEXP-902: (EST) OAS sync - Week 20

asein-sinch:
Breaking change introduced by the API: there are now 2 variations of a SipEndpoint (static or registered) => as there is no discriminator, Typescript doesn't know which Type to use and a guard needs to be used to access type-specific properties.


JPPortier:
> as there is no

Right; this is valid from a spec of view but not helpful from "consumer": need to try each of de-serialization regarding required fields existence to interpret the schema.

JPPortier (approved):

PR #177, opened 2025-07-01
DEVEXP-1035: Retrofit fix refresh token in next version

asein-sinch:


packages/sdk-client/src/api/api-errors.ts
```diff
@@ -62,15 +62,14 @@ export class RequestFailedError<T> extends GenericError {
...
  */
-export class EmptyResponseError<T> extends GenericError {
+export class EmptyResponseError extends GenericError {
   /**
    * Data decoded from the response body
    */
   public data?: string;
 
-  constructor(message: string, errorContext: ErrorContext, data?: T) {
+  constructor(message: string, errorContext: ErrorContext) {
```
JPPortier, line 71: For an empty response, we could safely remove this class level and unused `data` field

PR #179, opened 2025-07-02
DEVEXP-1005: (Fax + Numbers) OAS sync - Week 25

asein-sinch:


packages/numbers/src/models/v1/requests/available-regions/available-regions-request-data.ts
```diff
@@ -1,11 +1,15 @@
+import { NumberTypeEnum } from '../../enums';
+
+/** @deprecated Use `NumberTypeEnum` instead. */
```
JPPortier, line 3: ❤️

JPPortier (approved):

PR #191, opened 2025-08-27
DEVEXP-1092: (Conversation) OAS sync - Week 35

asein-sinch:


packages/conversation/tests/rest/v1/transcoding/transcoding.steps.ts
```diff
@@ -41,7 +41,7 @@ When('I send a request to transcode a location message', async () => {
         'RCS',
         'SMS',
         'TELEGRAM',
-        'VIBER',
+        'VIBER', // Kept for backward compatibility
```
JPPortier, line 44: Why not updating e2e ?

asein-sinch: Because the API still works and the response data in the mock is still valid

JPPortier (approved):

PR #192, opened 2025-08-27
DEVEXP-1018: (Verification API) Snippets

asein-sinch:


docs/snippets/verification/report/report-sms-by-id.js
```diff
@@ -0,0 +1,28 @@
...
+import { SinchClient } from '@sinch/sdk-core';
+import * as dotenv from 'dotenv';
+dotenv.config();
+
+(async () => {
+  const applicationKey = process.env.SINCH_APPLICATION_API_KEY || 'MY_APP_KEY_ID';
+  const applicationSecret = process.env.SINCH_APPLICATION_API_SECRET || 'MY_APP_KEY_SECRET';
+
+  const verificationId = 'THE_VERIFICATION_ID';
+  const receivedVerificationCode = 'THE_RECEIVED_VERIFICATION_CODE';
```
JPPortier, line 14: We should follow a same naming convention across SDKs (cf https://github.com/sinch/sinch-sdk-java-snippets/blob/main/snippets/src/main/java/verification/report/ReportById.java)
- at snippet level: it do not hurt, but at documentation level, switching from one snippet to another won't help 
- if documentation team want to provide "general" introduction help, they won't be able to

docs/snippets/verification/report/report-flashcall-by-id.js
```diff
@@ -6,23 +6,32 @@ import { SinchClient } from '@sinch/sdk-core';
 import * as dotenv from 'dotenv';
 dotenv.config();
 
-(async () => {
-  const applicationKey = process.env.SINCH_APPLICATION_API_KEY || 'MY_APP_KEY_ID';
-  const applicationSecret = process.env.SINCH_APPLICATION_API_SECRET || 'MY_APP_KEY_SECRET';
+async function main() {
+  const applicationKey = process.env.SINCH_APPLICATION_API_KEY ?? 'MY_APP_KEY_ID';
+  const applicationSecret = process.env.SINCH_APPLICATION_API_SECRET ?? 'MY_APP_KEY_SECRET';
```
JPPortier, line 11: What do you think about using same naming in use from dashboard and SDKs  ? (https://github.com/sinch/sinch-sdk-node?tab=readme-ov-file#constructor)

```suggestion
  const applicationKey = process.env.SINCH_APPLICATION_KEY ?? 'MY_APPLICATION_KEY';
  const applicationSecret = process.env.SINCH_APPLICATION_SECRET ?? 'APPLICATION_SECRET';
```

docs/snippets/verification/status/get-by-identity.js
```diff
@@ -6,18 +6,27 @@ import { SinchClient } from '@sinch/sdk-core';
...
   const sinch = new SinchClient({ applicationKey, applicationSecret });
 
-  const response = await sinch.verification.verificationStatus.getByIdentity({
-    endpoint: phoneNumber,
-    method: 'sms',
-  });
+  try {
+    const response = await sinch.verification.verificationStatus.getByIdentity({
+      endpoint: phoneNumber,
+      method: 'sms',
```
JPPortier, line 22: We could define a variable (like  for `phoneNumber`) at the beginning to highlight we can perform a getByIdentity onto other method.
Here, the hardcoded `sms` is mixin varibale input conditions with constants ones (but the constants is not required as for an identity `number` parameter onto a "start")
Added value: for any user running. this snippet: all input parameters are located at same  place: if a "start" was performed onto whatsapp, this mean the this snippet will fail if only phone number value is changed

asein-sinch: Fixed (and also in the SDK, the non-enum string related to the verification method)

JPPortier (approved):

PR #219, opened 2025-12-02
Bugfix: Add debug option to log response headers

asein-sinch:
**How to enable debug logs:**
```
import winston from 'winston';
const winstonLogger = winston.createLogger({
  level: 'debug',
  transports: [new winston.transports.Console()],
  format: winston.format.logstash(),
});

new SinchClient({ projectId, keyId, keySecret, logger:  winstonLogger});
```

**Debug logs:**
```
[Sinch SDK][Debug][AvailableRegionsApi][ListAvailableRegions][401]
HTTP method: GET
URL: https://numbers.api.sinch.com/v1/projects/37b62a7b-0177-429a-bb0b-e10f848de0b8/availableRegions?types=LOCAL&types=MOBILE
Response Headers:  Headers {
  [Symbol(map)]: [Object: null prototype] {
    'www-authenticate': [
      'Bearer error="invalid_token", error_description="Jwt expired at 2023-10-02T14:57:08Z", error_uri="https://tools.ietf.org/html/rfc6750#section-3.1"'
    ],
    'cache-control': [ 'no-cache, no-store, max-age=0, must-revalidate' ],
    pragma: [ 'no-cache' ],
    expires: [ '0' ],
    'x-content-type-options': [ 'nosniff' ],
    'strict-transport-security': [ 'max-age=31536000 ; includeSubDomains' ],
    'x-frame-options': [ 'DENY' ],
    'x-xss-protection': [ '0' ],
    'referrer-policy': [ 'no-referrer' ],
    'x-envoy-upstream-service-time': [ '2' ],
    date: [ 'Tue, 02 Dec 2025 22:25:38 GMT' ],
    server: [ 'istio-envoy' ],
    connection: [ 'close' ],
    'content-length': [ '0' ]
  }
}

```

### DIFF
--- a/examples/snippets/sdk-client/logger.js
+++ b/examples/snippets/sdk-client/logger.js
@@ -1,11 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
 /**
  * Sinch Node.js Snippet
  * See: https://github.com/sinch/sinch-sdk-node/examples/snippets
  *
  * The SDK accepts any object implementing the Logger interface (debug, info, warn, error).
- * Pass it via the `logger` option on SinchClient.
+ * Pass it via the `logger` option on SinchClient as it is wrapped in SinchLogger automatically.
  *
- * Example with Winston (install separately: npm install winston):
+ * Example with Winston (https://www.npmjs.com/package/winston or npm install winston):
  *
  *   import winston from 'winston';
  *
@@ -20,22 +22,30 @@
 import { SinchClient } from '@sinch/sdk-core';
 import * as dotenv from 'dotenv';
 
+
 dotenv.config();
+
+const appLogger = {
+  debug: (message, ...meta) => console.debug(message, ...meta),
+  info: (message, ...meta) => console.info(message, ...meta),
+  warn: (message, ...meta) => console.warn(message, ...meta),
+  error: (message, ...meta) => console.error(message, ...meta),
+};
 
 async function main() {
   const projectId = process.env.SINCH_PROJECT_ID ?? 'MY_PROJECT_ID';
   const keyId = process.env.SINCH_KEY_ID ?? 'MY_KEY_ID';
   const keySecret = process.env.SINCH_KEY_SECRET ?? 'MY_KEY_SECRET';
 
-  // Omit `logger` to use console, pass `logger: null` to silence SDK output,
-  // or pass a custom logger — see the Winston example in the header comment.
-  const sinch = new SinchClient({ projectId, keyId, keySecret });
+  // conversationRegion is intentionally omitted to trigger an SDK deprecation warning
+  // and show the custom logger in action
+  const sinch = new SinchClient({ projectId, keyId, keySecret, logger: appLogger });
 
   try {
-    await sinch.numbers.availableRegions.list();
+    await sinch.conversation.app.list({});
     console.log('✅ SDK request completed.');
   } catch (err) {
-    console.error('❌ Failed to call the Numbers API:');
+    console.error('❌ Failed to call the Conversation API:');
     console.error(err);
   }
 }
--- a/examples/snippets/sdk-client/logger.js
+++ b/examples/snippets/sdk-client/logger.js
@@ -1,0 +1,43 @@
+/**
+ * Sinch Node.js Snippet
+ * See: https://github.com/sinch/sinch-sdk-node/examples/snippets
+ *
+ * The SDK accepts any object implementing the Logger interface (debug, info, warn, error).
+ * Pass it via the `logger` option on SinchClient.
+ *
+ * Example with Winston (install separately: npm install winston):
+ *
+ *   import winston from 'winston';
+ *
+ *   const winstonLogger = winston.createLogger({
+ *     level: 'debug',
+ *     transports: [new winston.transports.Console()],
+ *     format: winston.format.logstash(),
+ *   });
+ *
+ *   new SinchClient({ projectId, keyId, keySecret, logger: winstonLogger });
+ */
+import { SinchClient } from '@sinch/sdk-core';
+import * as dotenv from 'dotenv';
+
+dotenv.config();
+
+async function main() {
+  const projectId = process.env.SINCH_PROJECT_ID ?? 'MY_PROJECT_ID';
+  const keyId = process.env.SINCH_KEY_ID ?? 'MY_KEY_ID';
+  const keySecret = process.env.SINCH_KEY_SECRET ?? 'MY_KEY_SECRET';
+
+  // Omit `logger` to use console, pass `logger: null` to silence SDK output,
+  // or pass a custom logger — see the Winston example in the header comment.
+  const sinch = new SinchClient({ projectId, keyId, keySecret });
+
+  try {
+    await sinch.numbers.availableRegions.list();
+    console.log('✅ SDK request completed.');
+  } catch (err) {
+    console.error('❌ Failed to call the Numbers API:');
+    console.error(err);
+  }
+}
+
+main();
--- a/examples/snippets/sdk-client/logger.js
+++ b/examples/snippets/sdk-client/logger.js
@@ -1,52 +1,43 @@
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-nocheck
 /**
  * Sinch Node.js Snippet
  * See: https://github.com/sinch/sinch-sdk-node/examples/snippets
  *
  * The SDK accepts any object implementing the Logger interface (debug, info, warn, error).
- * Pass it via the `logger` option on SinchClient as it is wrapped in SinchLogger automatically.
+ * Pass it via the `logger` option — it is wrapped in SinchLogger automatically.
  *
- * Example with Winston (https://www.npmjs.com/package/winston or npm install winston):
- *
- *   import winston from 'winston';
- *
- *   const winstonLogger = winston.createLogger({
- *     level: 'debug',
- *     transports: [new winston.transports.Console()],
- *     format: winston.format.logstash(),
- *   });
- *
- *   new SinchClient({ projectId, keyId, keySecret, logger: winstonLogger });
+ * This example uses Winston (https://www.npmjs.com/package/winston). An invalid OAuth
+ * endpoint is used on purpose so the token request fails with an HTTP error and the SDK
+ * logs the response details at debug level.
  */
-import { SinchClient } from '@sinch/sdk-core';
-import * as dotenv from 'dotenv';
+import { Oauth2TokenRequest } from '@sinch/sdk-core';
+import { Headers } from 'node-fetch';
+import * as winston from 'winston';
 
-
-dotenv.config();
-
-const appLogger = {
-  debug: (message, ...meta) => console.debug(message, ...meta),
-  info: (message, ...meta) => console.info(message, ...meta),
-  warn: (message, ...meta) => console.warn(message, ...meta),
-  error: (message, ...meta) => console.error(message, ...meta),
-};
+const winstonLogger = winston.createLogger({
+  level: 'debug',
+  transports: [new winston.transports.Console()],
+  format: winston.format.printf(({ level, message }) => {
+    const text = typeof message === 'function' ? message() : message;
+    return JSON.stringify({ '@fields': { level }, '@message': text });
+  }),
+});
 
 async function main() {
-  const projectId = process.env.SINCH_PROJECT_ID ?? 'MY_PROJECT_ID';
-  const keyId = process.env.SINCH_KEY_ID ?? 'MY_KEY_ID';
-  const keySecret = process.env.SINCH_KEY_SECRET ?? 'MY_KEY_SECRET';
+  const keyId = 'MY_KEY_ID';
+  const keySecret = 'MY_KEY_SECRET';
+  const invalidAuthHostname = 'https://example.com';
 
-  // conversationRegion is intentionally omitted to trigger an SDK deprecation warning
-  // and show the custom logger in action
-  const sinch = new SinchClient({ projectId, keyId, keySecret, logger: appLogger });
+  const oauthPlugin = new Oauth2TokenRequest(keyId, keySecret, invalidAuthHostname, winstonLogger);
 
   try {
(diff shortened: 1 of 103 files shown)


packages/sdk-client/src/plugins/exception/exception.response.ts
```diff
@@ -75,11 +80,25 @@ export class ExceptionResponse<
   }
 
   private debug(context: ResponsePluginContext) {
-    if (context.requestOptions.logHeadersOnError && !context.response?.ok) {
-      console.debug(
-        `[Sinch SDK][Debug][${context.apiName}][${context.operationId}][${context.response?.status}]\nHTTP method: ${context.requestOptions.method}\nURL: ${context.url}\nResponse Headers: `,
-        context.response?.headers,
+    if (!context.response?.ok) {
+      this.logger.debug(
+        `[Sinch SDK][Debug][${context.apiName}][${context.operationId}][${context.response?.status}]\nHTTP method: ${context.requestOptions.method}\nURL: ${context.url}\nResponse Headers: ${this.formatHeaders(context.response?.headers)}`,
```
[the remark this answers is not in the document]
asein-sinch: Done

packages/sdk-client/src/logger/logger.ts
JPPortier: ❤️ 
May be adding some unit tests to ensure proper/format and log level.
Not sure we want to go to this direction for console logger and this PR,  but other logging frameworks (e.g.: winston pointed by PR comment) are enabling to configure log level to not flood logs output/destination

JPPortier (approved):

packages/sms/src/rest/v1/sms-domain-api.ts
```diff
@@ -59,7 +60,9 @@ export class SmsDomainApi implements Api {
     try {
       this.lazyClient.getApiClient();
     } catch (error) {
-      console.error('Impossible to assign the new credentials to the SMS API');
+      new SinchLogger(this.lazyClient.sharedConfig.logger ?? console).error(
```
JPPortier: Here we are assuming we have to send a log onto console even if configuration could have been, by purpose of SDK user, to NOT have any logger (multliple times)

What do you think :
- either we assume there is, by contract, a logger defined
- test if null/undefined and do not log anything

packages/sdk-core/src/sinch-client.ts
```diff
@@ -26,6 +26,9 @@ export class SinchClient {
    * @param {SinchClientParameters} params - The object containing the Sinch credentials.
    */
   constructor(params: SinchClientParameters) {
+    if (!params.logger) {
```
JPPortier: See comment about using or not a default logger.
What if for any purpose an end user do not want any log ? (thinking about a logging call causing an issue)
It won't be possible.
We could have distinction between `null` VS `undefined` to provide 
- null -> a dummy logger not logging anything
- undefined -> a default logger based onto init of SinchLogger

And, by contract we could remove all validation against not defined logger. e.g.: https://github.com/sinch/sinch-sdk-node/pull/219/changes#diff-85a0299065d0f2bd71e5151e1983edb3447d2d9770692d9f8ac007467246bc20R69

packages/sdk-client/src/plugins/exception/exception.response.ts
```diff
@@ -9,18 +10,27 @@ export class ExceptionResponse<
...
+  ) {
+    this.logger = new SinchLogger(logger);
+  }
 
   public load(
     context: ResponsePluginContext,
   ): PluginRunner<V | Record<string, unknown>, V> {
     return {
       transform: (res: V) => {
+        this.debug(context);
```
JPPortier: 🤔
Not a big fan to have a dependency onto logger/SinchLogger here.
Finally: it is "just" a container.

We could think about who should be responsible of logging. 
Here, not sure a "simple" container should log itself something.
It is an exception response. Why not having transport layer performing the log itself and removing dependency onto any logger here?

packages/sdk-client/src/logger/logger.ts
```diff
@@ -0,0 +1,27 @@
+export interface Logger {
+  debug(message: string, ...meta: any[]): void;
+  info(message: string, ...meta: any[]): void;
+  warn(message: string, ...meta: any[]): void;
+  error(message: string, ...meta: any[]): void;
+}
```
JPPortier: We could imagine, for performance/memory consumption purpose having interface defining a callback called by logger when the message will be really send.
Then; it will avoid to format message before calling logger which will finally decide to not log anything because of required severity is not reached.

packages/sdk-client/src/domain/domain-interface.ts
```diff
@@ -186,3 +188,8 @@ export type MailgunRegion = SupportedMailgunRegion | string;
 export const MailgunRegion = {
   ...SupportedMailgunRegion,
 };
+
+export interface LoggerParameters {
```
JPPortier: Strictly speaking; this is nor defining "logger parameters" but the logger itself

packages/sms/src/rest/v1/sms-domain-api.ts
Hurus111: Fixed 🟡 
Removed all ?? console fallbacks; call sites use resolveLogger(...). logger: null → silent NOOP_LOGGER.
[packages/sms/src/rest/v1/sms-domain-api.ts](https://github.com/sinch/sinch-sdk-node/pull/219/files) (and all other *-service.ts / *-domain-api.ts)

packages/sdk-core/src/sinch-client.ts
Hurus111: Fixed 🟡 
- null -> silent
- undefined -> default console

[packages/sdk-core/src/sinch-client.ts](https://github.com/sinch/sinch-sdk-node/pull/219/files) resolved once via resolveLogger() in SinchClient and each Service constructor.

packages/sdk-client/src/plugins/exception/exception.response.ts
Hurus111: Fixed 🟡 
Logger removed from ExceptionResponse. Failed HTTP debug logging moved to ApiFetchClient.
[packages/sdk-client/src/client/api-fetch-client.ts](https://github.com/sinch/sinch-sdk-node/pull/219/files)

packages/sdk-client/src/domain/domain-interface.ts
Hurus111: Fixed 🟡 
Renamed to `WithLogger`

packages/sdk-client/src/logger/logger.ts
Hurus111: Done 🟢 
Logger methods now accept either a plain `string` or a `callback () => string`. The SDK builds the message only when the logger actually emits it.

examples/snippets/sdk-client/logger.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Sinch Node.js Snippet
+ * See: https://github.com/sinch/sinch-sdk-node/examples/snippets
+ *
+ * The SDK accepts any object implementing the Logger interface (debug, info, warn, error).
+ * Pass it via the `logger` option on SinchClient.
+ *
+ * Example with Winston (install separately: npm install winston):
```
JPPortier: Adding a link onto Winston npmjs package or github repo ?

examples/snippets/sdk-client/logger.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ *   import winston from 'winston';
+ *
+ *   const winstonLogger = winston.createLogger({
+ *     level: 'debug',
+ *     transports: [new winston.transports.Console()],
+ *     format: winston.format.logstash(),
+ *   });
+ *
```
JPPortier: Why not having this comment part of the snippet source itself ?
The snippet is to highlight how to add a custom snippet but finally we just init a client without any below

examples/snippets/sdk-client/logger.js
```diff
@@ -0,0 +1,43 @@
...
+dotenv.config();
+
+async function main() {
+  const projectId = process.env.SINCH_PROJECT_ID ?? 'MY_PROJECT_ID';
+  const keyId = process.env.SINCH_KEY_ID ?? 'MY_KEY_ID';
+  const keySecret = process.env.SINCH_KEY_SECRET ?? 'MY_KEY_SECRET';
+
+  // Omit `logger` to use console, pass `logger: null` to silence SDK output,
+  // or pass a custom logger — see the Winston example in the header comment.
+  const sinch = new SinchClient({ projectId, keyId, keySecret });
```
JPPortier: Provide here (see comment above) an init with a defined logger

packages/conversation/src/rest/v1/conversation-domain-api.ts
```diff
@@ -67,7 +65,7 @@ export class ConversationDomainApi implements Api {
     try {
       this.lazyClient.getApiClient();
     } catch (error) {
-      new SinchLogger(resolveLogger(this.lazyClient.sharedConfig.logger)).error(
+      this.lazyClient.sharedConfig.logger!.error(
```
JPPortier: We should never receive a null/undefined `this.lazyClient.sharedConfig.logger` here.
If sinch client init is checking against:
- null: then a `NOOP_LOGGER` should have been created and in use 
- undefined: then a `CONSOLE_LOGGER` should have been created and in use

packages/numbers/src/rest/v1/numbers-service.ts
```diff
@@ -76,6 +77,7 @@ export class NumbersService {
    * @param {SinchClientParameters} params - an Object containing the necessary properties to initialize the service
    */
   constructor(params: SinchClientParameters) {
+    params.logger = resolveLogger(params.logger);
```
JPPortier: Here the input `params` parameter is modified itself.
Better to store result of resolvConf wihtin `sharedClient` right ?
(could be the cause of these errors: https://github.com/sinch/sinch-sdk-node/actions/runs/27703589896/job/81946190935)

examples/snippets/sdk-client/logger.js
Hurus111: I wanted to avoid adding more dependencies to the project (even though it's only used here) @JPPortier

examples/snippets/sdk-client/logger.js
Hurus111: Added ` * Example with Winston (https://www.npmjs.com/package/winston — install separately: npm install winston):
`

packages/conversation/src/rest/v1/conversation-domain-api.ts
Hurus111: After init, `sharedConfig.logger` is always a resolved Logger via `ResolvedSinchClientParameters`. So no more `logger!` everywhere

packages/numbers/src/rest/v1/numbers-service.ts
Hurus111: The service no longer mutates caller params; it uses `resolveClientParameters` and stores the result on the `lazyClient.sharedConfig`

examples/snippets/sdk-client/logger.js
Hurus111: Agreed the snippet should pass a logger in code: we'll use a minimal inline implementation so it runs without extra dependecies. If it's fine for you, Winston can stay in the comment as an optional integration example

Hurus111: Also added intentionally deprecated call to show custom log:
> [Sinch SDK][Warn] ** DEPRECATION NOTICE ** The "conversationRegion" property will become mandatory in the next major version of the SDK and not default to "us" anymore. Please set it to a valid region.
✅ SDK request completed.

packages/numbers/src/rest/v1/numbers-service.ts
Hurus111: I also added a shared `LazyApiClient` base class in `@sinch/sdk-client` so each domain lazy client only implements `getApiClient()`

packages/sdk-client/src/client/lazy-api-client.ts
```diff
@@ -0,0 +1,17 @@
+import { ResolvedSinchClientParameters, SinchClientParameters } from '../domain';
+import { resolveClientParameters } from '../logger';
+import { ApiFetchClient } from './api-fetch-client';
+
+/** Base class for domain lazy clients that share resolved SDK configuration. */
+export abstract class LazyApiClient {
+  public apiFetchClient?: ApiFetchClient;
+  public sharedConfig: ResolvedSinchClientParameters;
+
+  constructor(params: SinchClientParameters | ResolvedSinchClientParameters) {
```
JPPortier: Is the `SinchClientParameters |` is still required ?
Regarding the service files, they perform a call to `const sharedConfig = resolveClientParameters(params);` before creating the underlying client, so the LazyClient will always receive a `ResolvedSinchClientParameters` right ?
e.g. from : https://github.com/sinch/sinch-sdk-node/pull/219/changes#diff-e08475f2da0aa9048bf9aaf2fa12ce8a08638fb91dbdec846c1276355d3905dbR127

packages/conversation/src/rest/v1/conversation-service.ts
```diff
@@ -132,10 +124,11 @@ export class ConversationService {
    * @param {SinchClientParameters} params - an Object containing the necessary properties to initialize the service
    */
   constructor(params: SinchClientParameters) {
-    const sharedConversationClient = new LazyConversationApiClient(params);
+    const sharedConfig = resolveClientParameters(params);
```
JPPortier: In `setApiClientConfig` function just below, the variable is named `resolvedParams`
Better to have a common pattern/naming

packages/elastic-sip-trunking/src/rest/v1/elastic-sip-trunking-service.ts
JPPortier, line 1: ConversationService (all all other services) are having a public `setApiClientConfig` function https://github.com/sinch/sinch-sdk-node/blob/4cf6337c9748afc78a4bfd3e6073516c917fa4ab/packages/conversation/src/rest/v1/conversation-service.ts#L148C10-L148C28 and a setCredentials.
Here only setCredentials. 
Why this special use case for EST ?

JPPortier (approved):

PR #241, opened 2026-04-12
Handle multipart/form-data webhooks

asein-sinch:


packages/fax/src/rest/v3/callbacks/callbacks-webhook.ts
```diff
@@ -8,11 +8,19 @@ export class FaxCallbackWebhooks implements CallbackProcessor<FaxWebhookEventPar
...
+   *   - A JSON string (application/json content type)
+   *   - A multipart/form-data raw body string (fields will be extracted automatically)
+   *   - An already-parsed object
    * @return {FaxWebhookEventParsed} - The parsed Fax event object
    */
   public parseEvent(eventBody: any): FaxWebhookEventParsed {
     if (typeof eventBody === 'string') {
-      eventBody = JSON.parse(eventBody);
+      if (eventBody.trimStart().startsWith('--')) {
+        eventBody = this.parseMultipartFormData(eventBody);
```
rpredescu-sinch, line 20: Hi, @asein-sinch!

Should we trim `eventBody` before passing it to `parseMultipartFormData()`? The ^ regex inside the method will fail to match if the body has leading whitespace, even though the `trimStart()` check in the condition already handles it. Passing the trimmed string keeps the two in sync:

- `eventBody = this.parseMultipartFormData(eventBody.trimStart());`

What do you think?

packages/fax/tests/rest/v3/webhooks/callbacks-webhook.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      fax: { ...FAX_OBJECT },
+    });
+    const result = callbackWebhooks.parseEvent(payload) as Fax.FaxCompletedEvent;
+    expect(result.event).toBe('FAX_COMPLETED');
+    expect(result.eventTime).toStrictEqual(EVENT_TIME_AS_DATE);
+    expect(result.fax!.id).toBe('01KNPQW6SGZKAZEX8ZX7MCN559');
+    expect(result.fax!.createTime).toStrictEqual(CREATE_TIME_AS_DATE);
+  });
+
+  it('should parse an INCOMING_FAX event from a multipart/form-data body with CRLF', () => {
```
rpredescu-sinch, line 84: Hi, @asein-sinch!
We have the **CRLF** tests for the multipart parsing. Is there any value to add equivalent ones for **LF**-only line endings `(.join('\n')`)?
Best!

packages/fax/src/rest/v3/callbacks/callbacks-webhook.ts
```diff
@@ -48,7 +48,6 @@ export class FaxCallbackWebhooks implements CallbackProcessor<FaxWebhookEventPar
           throw new Error(`Unknown Fax event: ${eventBody.event}`);
       }
     }
-    console.log(eventBody);
     throw new Error('Unknown Fax event');
```
asein-sinch: Since you are removing the console.log,add the stringified version of the event body in the error, it's better to report some details to the user.
And do the same for voice and verification events

Hurus111 (approved):
Looks good 😉

PR #242, opened 2026-04-13
Bugfix: make lazyClient public

asein-sinch:


packages/conversation/src/rest/v1/conversation-service.ts
```diff
@@ -23,7 +23,7 @@ import { ConsentsApi } from './consents';
 import { DEFAULT_CONVERSATION_REGION_DEPRECATION_WARNING } from './conversation-domain-api';
 
 export class LazyConversationApiClient {
-  private apiFetchClient?: ApiFetchClient;
+  apiFetchClient?: ApiFetchClient;
```
matsk-sinch, line 26: If you omit public, it assumes as `public`?

packages/voice/tests/rest/v1/voice-service.test.ts
```diff
@@ -178,4 +191,56 @@ describe('Voice Service', () => {
     expect((voiceService as any).lazyVoiceClient.sharedConfig.applicationKey).toBe('APPLICATION_KEY');
```
matsk-sinch: these `as any` casts are needed, since it's public?

packages/conversation/src/rest/v1/conversation-service.ts
asein-sinch: Yes exactly

packages/voice/tests/rest/v1/voice-service.test.ts
asein-sinch: You're right, they are not needed anymore. I've removed them for all the APIs

matsk-sinch (approved):

PR #244, opened 2026-05-11
DEVEXP-1388: Refresh token harness

asein-sinch:


packages/sdk-client/src/plugins/oauth2/oauth2-token.request.ts
```diff
@@ -107,17 +110,43 @@ export class Oauth2TokenRequest implements RequestPlugin {
...
 
 enum TokenStatus {
   INVALID,
   VALID,
 }
+
+/**
+ * Decode the `exp` claim from a JWT without verifying the signature.
+ */
+const decodeJwtExp = (jwt: string): number | undefined => {
```
Hurus111, line 140: Why not go for a JWT decoding library? I get that for this case only it's perfectly reasonable to just have this manual decoder, and you also reduce dependencies, but did you consider adding a lib? Maybe for future implementations it'd be better to have a lib for this 🤔

packages/sdk-client/src/plugins/oauth2/oauth2-token.request.ts
```diff
@@ -107,17 +110,43 @@ export class Oauth2TokenRequest implements RequestPlugin {
...
+   * Invalidate the cached token. If `failingJwt` is supplied, the cache is only
+   * cleared when it still holds that exact JWT. This prevents a stale-JWT 401
+   * retry from wiping a token that another caller already successfully refreshed.
+   */
+  public invalidateToken(failingJwt?: string) {
+    if (failingJwt !== undefined && this.token && this.token.value !== failingJwt) {
+      return;
+    }
     this.token = undefined;
   }
```
Hurus111: Rename **invalidateToken** to **clearCachedToken** so the name matches the behavior: we only drop the in-memory OAuth2 bearer, and when a failing access token is passed we clear only if the cache still holds the same value.

Hurus111 (commented):
Awesome! Just a couple of questions/propositions 😄

packages/sdk-client/src/plugins/oauth2/oauth2-token.request.ts
asein-sinch: Precisely because one of the SDK requirement is to have 0 external dependencies. We'll see how much this requirement makes sense in the future, but for this chirurgical operation to read a single property in the JWT, this is fine to keep it like this

packages/sdk-client/src/plugins/oauth2/oauth2-token.request.ts
asein-sinch: Good remark, thanks. I've fixed the method name and the related comments

Hurus111 (approved):

PR #249, opened 2026-05-18
fix(fax): download content - set fileFormat as deprecated

Hurus111:


asein-sinch (approved):

asein-sinch:
Normally, this should have gone inside the `v1.5-next` branch but let's ship it as a bug fix, along with the PR https://github.com/sinch/sinch-sdk-node/pull/241

PR #252, opened 2026-06-01
DEVEXP-1401: (Voice) OAS sync - Sprint 8 bis

Hurus111:
@asein-sinch `SVAMLRequestBody` was removed/replaced in the API registry, so I wasn't sure whether to drop it from the SDK too or keep it around.

I decided to deprecate it instead. It gives existing consumers a migration path (`ManagedCallSvamlRequestBody` for call management, `AceResponse` / `IceResponse` / `PieResponse` for callbacks) without an extra breaking change on top of the stricter types on `updateCall` and `manageCallWithCallLeg`. 

We can remove it in a future update (or keep it deprecated). What do you think?


packages/voice/src/models/v1/svaml-request-body/svaml-request-body.ts
```diff
@@ -1,7 +1,36 @@
...
+  SvamlAction,
+  SvamlActionContinue,
+  SvamlActionHangup,
+  SvamlActionPark,
+  SvamlInstruction,
+} from '../mod-svaml';
+
+/**
+ * SVAML actions allowed when managing an ongoing call via
+ * Connect and runMenu actions are not supported on these endpoints.
```
asein-sinch: Review this description, it seems some words are missing

packages/voice/tests/rest/v1/calls/calls-api.test.ts
```diff
@@ -67,7 +67,7 @@ describe('CallsApi', () => {
         text: 'Hello, the call is over, hanging up now. Goodbye',
         locale: 'en-US',
       } as Voice.SvamlInstructionSay;
-      const action: Voice.SvamlAction = {
+      const action: Voice.ManagedCallSvamlAction = {
         name: 'hangup',
       } as Voice.SvamlActionHangup;
```
asein-sinch, line 72: Happy to see that the test was using the appropriate action already 😮‍💨

packages/voice/src/models/v1/svaml-request-body/svaml-request-body.ts
```diff
@@ -1,7 +1,36 @@
...
+
+  /** The collection of instructions that can perform various tasks during the call. You can include as many instructions as necessary. */
+  instructions?: SvamlInstruction[];
+  /** The action that will control the call. Each SVAML object can only include one action. */
+  action?: ManagedCallSvamlAction;
+}
 
 /**
  * SVAML is a call control markup language. When a server receives a callback event from the Sinch platform, it can respond with a SVAML object to control the voice call. The following is an example of a SVAML object type and its contents.
+ * @deprecated SVAML is context-specific. Use {@link ManagedCallSvamlRequestBody} for
```
asein-sinch, line 28: Yes, deprecation is the way to go, no breaking change is introduced in this version. We can remove it in the next major if there is no strong signal from customers to keep it.

packages/voice/src/models/v1/svaml-request-body/svaml-request-body.ts
```diff
@@ -6,17 +6,12 @@ import {
...
 export type ManagedCallSvamlAction = SvamlActionHangup
   | SvamlActionContinue
   | SvamlActionPark;
 
 /**
- * SVAML request body for managing ongoing, connected calls.
- * Only hangup, continue, and park actions are valid on these endpoints.
+ * SVAML is a call control markup language. When a server receives a callback event from the Sinch platform, it can respond with a SVAML object to control the voice call. The following is an example of a SVAML object type and its contents.
  */
 export interface ManagedCallSvamlRequestBody {
```
asein-sinch: In fact the previous comments were better.

- For `ManagedCallSvamlAction`, you can keep "SVAML actions allowed when managing an ongoing call"
- For `ManagedCallSvamlRequestBody`, the 2 first lines were better. The one from the spec that is used on the deprecated schema is not very explicative.

asein-sinch (approved):

PR #256, opened 2026-06-10
DEVEXP-1409: README unification

Hurus111:


README.md
```diff
@@ -1,180 +1,399 @@
...
+- [Third-party dependencies](#third-party-dependencies)
+- [Examples](#examples)
+- [Changelog](#changelog)
+- [License](#license)
+- [Contact](#contact)
 
 ## Prerequisites
 
-Before being able to execute the commands described below, you will need to install Node.js. We recommend [installing the current or the LTS version](https://nodejs.org/en).
+- Node.js 18 or later (LTS recommended)
```
JPPortier: Can we provide exact list of version used from CI matrix for validation ?
Do we have HTTP link onto each of them ?

README.md
```diff
@@ -1,180 +1,399 @@
...
+yarn init
+yarn add @sinch/sdk-core
 ```
 
-> **Warning**:
-> Do not use this Node.js library in front-end applications (such as Angular, React, Vue.js, etc.). Doing so can expose your Sinch credentials to end-users as part of the HTML/JavaScript files loaded on their browser.
+## Supported APIs
 
-### Node.js supported versions
+| API Category    | API Name |
```
JPPortier, line 54: Can we have empty cells onto 1st column when API category is same across rows ?  (@marcos-sinch same for Python)

README.md
```diff
@@ -1,180 +1,399 @@
...
+|-----------------|----------|
+| Messaging       | [Conversation API](https://developers.sinch.com/docs/conversation) |
+| Messaging       | [SMS API](https://developers.sinch.com/docs/sms) |
```
JPPortier: We could add also here a note about "End Of Sale" for SMS to avoid any new user going to implement SMS usage and not Conversation (@marcos-sinch same for Python)

README.md
```diff
@@ -1,180 +1,399 @@
...
+  return res.status(401).send('Invalid webhook signature');
+}
+
+const event = callbackWebhooks.parseEvent(request.body);
 ```
 
-## SDK usage
+`SINCH_CONVERSATION_APP_SECRET` is the app secret set per app in the [Conversation dashboard](https://dashboard.sinch.com/convapi/apps). `parseEvent` works without validating the request, but then its origin can't be verified, so validating is recommended in production.
+
+You can find a complete example in [examples/webhooks](./examples/webhooks).
```
JPPortier: May be pointing to the section related to Conversation: https://github.com/sinch/sinch-sdk-node/blob/af26fb00d9fa8c21dbd79bf43c9b1f9546f9ec3e/examples/webhooks/src/controller/app.controller.ts#L41-L61

JPPortier (approved):
Approved (but conflicts with main branch need to be fixed before merge)

PR #260, opened 2026-06-24
DEVEXP-1411--typedoc-api

Hurus111:
## Summary
- Add TypeDoc API reference under `docs/api/` with domain-based sidebar categories
- Filter internal/test symbols (fixtures, pagination helpers, sdk-client plugins) from published docs
- Build docs exclusively from monorepo local packages + zip file as output


docs/api/package.json
```diff
@@ -3,20 +3,14 @@
   "private": true,
   "description": "TypeDoc API reference for @sinch/sdk-core",
   "scripts": {
-    "sync-version": "node sync-version.mjs",
-    "prepare-readme": "node prepare-readme.mjs",
     "doc": "node prepare-readme.mjs && typedoc",
```
JPPortier: For other SDK we do not add (yet) README 
Better to not have it

JPPortier (approved):

